PR #424, opened 2023-05-18
fix: reset Copied text back to default value after a timeout

omairnabiel:
# PR Type

- [x] Bugfix
- [x] Other... Please describe:

Whenever Stacks or Bitcoin address is copied it shows Copied tooltip text. That text never gets reset. This aims to reset the text back to it's default value after a timeout

# What is the current behaviour?
When Stacks or Bitcoin address is copied it's tooltip shows Copied indefinitely

# What is the new behaviour?
As commonly observed across other applications where text gets reset, similarly reset the `Copied` text back to default value after 3 seconds

# Screenshot / Video

### DIFF
--- a/src/app/components/accountRow/index.tsx
+++ b/src/app/components/accountRow/index.tsx
@@ -9,7 +9,7 @@ import BarLoader from '@components/barLoader';
 import Copy from '@assets/img/Copy.svg';
 import { LoaderSize } from '@utils/constants';
 import { Account } from '@secretkeylabs/xverse-core';
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { useDispatch } from 'react-redux';
 import { ChangeShowBtcReceiveAlertAction } from '@stores/wallet/actions/actionCreators';
 import useWalletSelector from '@hooks/useWalletSelector';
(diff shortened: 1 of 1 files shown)


src/app/components/accountRow/index.tsx
```diff
@@ -162,13 +162,19 @@ function AccountRow({
     navigator.clipboard.writeText(account?.btcAddress!);
     setOnBtcCopied(true);
     setOnStxCopied(false);
-    if (showBtcReceiveAlert !== null) { dispatch(ChangeShowBtcReceiveAlertAction(true)); }
+    // set 'Copied' text back to 'Bitcoin address' after 3 seconds
+    setTimeout(() => setOnBtcCopied(false), 3000);
```
dhriaznov, line 166: We should also clear the timeout when the component unmounts.

src/app/components/accountRow/index.tsx
```diff
@@ -162,13 +162,19 @@ function AccountRow({
...
+      dispatch(ChangeShowBtcReceiveAlertAction(true));
+    }
   };
 
   const handleOnStxAddressClick = () => {
     navigator.clipboard.writeText(account?.stxAddress!);
     setOnStxCopied(true);
     setOnBtcCopied(false);
+    // set 'Copied' text back to 'Stacks address' after 3 seconds
+    setTimeout(() => setOnStxCopied(false), 3000);
```
dhriaznov, line 177: Same here

src/app/components/accountRow/index.tsx
omairnabiel: Added that in this commit https://github.com/secretkeylabs/xverse-web-extension/pull/424/commits/14ae91a94adebf2f61820dd7d5687f860732bc5d

dhriaznov:
Thanks, @omairnabiel !

Issue #406, opened 2023-05-13
Support simple BTC send request from Sats connect


yknl:
Allow apps to request sending of Bitcoin to a specified address through Sats connect.

PR #439, opened 2023-05-23
Allow send btc through sats connect

Imamah-Zafar:
# PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

(Optional) Resolved: #406 

# What is the new behavior?
Allows sending btc transaction through sats connect

# Screenshot / Video

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/3e163397-d44f-4cec-a028-66fa8e15f979

Incase amount is less than the acceptable amount

failed

# Testing Instructions 
Needs to be linked locally with the sats-connect PR: https://github.com/secretkeylabs/sats-connect/pull/8
The test app has also been updated in PR: https://github.com/secretkeylabs/sats-connect-example/pull/2


yknl:
Getting this error in the popup after seeing the spinner for a few seconds

image

Imamah-Zafar:
> Getting this error in the popup after seeing the spinner for a few seconds
> 
> image

can you let me know if you made any changes in the test app? I can try the same changes because currently I am unable to reproduce the crash. I tried with multiple accounts and wallet and the behaviour is as expected 

1st

DuskaT021:
Hey @yknl I wasn't able to replicate the issue you reported.

https://github.com/secretkeylabs/xverse-web-extension/assets/55587184/9cc88970-1728-455c-b479-6c46c1d60282

> Getting this error in the popup after seeing the spinner for a few seconds
> 
> image

chunwei:
Could you kindly provide an update on the status of the PR? It would be greatly appreciated as my app relies on this particular feature.

Imamah-Zafar:
Before the recent commit, if the user had insufficient balance this was being displayed
error-1

The latest commit, fixes this and a proper detailed error message is displayed 
error

yknl:
This error message is misleading. The transaction hasn't failed, the transaction simply cannot be constructed due to insufficient balance. The error message should be revised to: "Invalid Transaction", "The requested transaction cannot be created due to insufficient balance.

image

yknl:
Several issues here:

- Weird scrolling behaviour on the popup, buttons are partially hidden
- Clicking cancel should close the popup not go back to an input screen 

https://github.com/secretkeylabs/xverse-web-extension/assets/29081231/f5c6f7c3-5236-4f20-b25e-b3d1bee475bd

yknl:
It would be a better user experience to be able to select which account to send from if the app did not specify a from address.

We should add the option of specifying a from address in Sats connect.

And while we're making these changes, we should change the send BTC API to allow a list of recipients and amounts.

Imamah-Zafar:
Latest commit requires to run with the [sats connect PR ](https://github.com/secretkeylabs/sats-connect/pull/10) and  [test app PR](https://github.com/secretkeylabs/sats-connect-example/pull/2)

Includes multiple recipients and amount and also adds sender address . Also adds fix for above mentioned changes

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/4cb38408-bfd1-4eb4-a2c9-1a93a1eaad5a


invalid-tx

yknl (dismissed):

DuskaT021:
needs conflict resolved

DuskaT021:
These two pr's need to be reviewed that hold the latest fixes https://github.com/secretkeylabs/xverse-web-extension/pull/439#issuecomment-1592347494 mentioned in the comment

src/app/hooks/useSendBtcRequest.ts
```diff
@@ -0,0 +1,62 @@
...
+import { useLocation } from 'react-router-dom';
+import { SendBtcTransactionOptions } from 'sats-connect';
+import useWalletSelector from './useWalletSelector';
+
+function useSendBtcRequest() {
+  const { search } = useLocation();
+  const params = new URLSearchParams(search);
+  const [recipient, setRecipient] = useState<Recipient[]>([]);
+  const requestToken = params.get('sendBtcRequest') ?? '';
+  const request = decodeToken(requestToken) as any as SendBtcTransactionOptions;
```
teebszet, line 15: is the `as any` redundant here?

src/app/components/confirmBtcTransactionComponent/index.tsx
```diff
@@ -343,9 +350,9 @@ function ConfirmBtcTransactionComponent({
             recipients?.map((recipient, index) => (
               <RecipientComponent
                 recipientIndex={index + 1}
-                address={recipient?.address}
-                value={satsToBtc(recipient?.amountSats).toString()}
-                totalRecipient={recipients?.length}
```
teebszet, line 348: why do we remove the undefined checks here?

Imamah-Zafar: the undefined check was unnecessary here. The undefined check should be with the recipient array which is added. 
Here we iterate over the recipient array and every object in the array will have an address and amountSats since these parameters are not optional.

src/app/hooks/useSendBtcRequest.ts
Imamah-Zafar: Basically The first casting (as any) is used when the type of decodeToken function's return value is not known or incompatible with the desired type. It allows treating the result as an any type, which essentially removes any type checking.The second casting (as SendBtcTransactionOptions) is attempting to explicitly assign the any type to SendBtcTransactionOptions. We know that the requestToken is compatible with SendBtcTransactionOptions since we do type checking in the sats connect package. If any attribute of SendBtcTransactionOptions is missing then the sats-connect package will throw an error. 
This was added here to be consistent with other sats-connect function screens  like `signPsbtScreen` and `getAddressScreen` since we are using the same practice in those screens as well but from my understanding , this is okay to parse responses from browser in this way

teebszet: yeah ok. feels like there should be a way to type this correctly though

teebszet (dismissed):

yknl (approved):

DuskaT021:
left my comment in the [linear ticket ](https://linear.app/xverseapp/issue/ENG-2151#comment-4c92f454)

DuskaT021:
this can be merged and the related pr in the [sats-connect](https://github.com/secretkeylabs/sats-connect/pull/10)

teebszet (approved):

PR #457, opened 2023-06-05
fix for auto lock

m-aboelenein:
# PR Type

- [x] Bugfix
- [ ] Enhancement
- [ ] Build related changes
- [ ] CI related changes

the extension requires the user to enter his/her password too frequently 

# What is the new behavior?
the extension attempts to reload to prevent this behavior


yknl:
It seems the auto-lock time is based on the last time the wallet was unlocked. The timer isn't refreshed when the wallet is re-opened. Is this the intended behaviour? Would it be better to refresh the timer every time the wallet is opened?

yknl (approved):

PR #459, opened 2023-06-06
Fix network in psbt parsing

abdulhaseeb4239:
# PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Enhancement
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

# What is the current behavior?
Resolved:[ ENG-2227](https://linear.app/xverseapp/issue/ENG-2227/magic-eden-psbt-signing-testnet-issue)

# What is the new behavior?
addresses should show up fine in sign PSBT screen on testnet

# Screenshot / Video

PR #470, opened 2023-06-11
Fix PSBT signing testnet output address showing as mainnet

yknl:
# PR Type

- [x] Bugfix
- [ ] Enhancement
- [ ] Build related changes
- [ ] CI related changes

When creating a PSBT signing request, the network is not passed to the PSBT parser. This results in testnet addresses in the outputs being parsed and displayed as mainnet addresses. And throws off the BTC amount transferred calculation because the user's own address is not recognized.

This bug only appears when in testnet mode.

# What is the new behavior?
Network is properly passed and addresses will display for the correct network.

# Screenshot / Video
Before:
image

After:
image

### DIFF
--- a/src/app/screens/signPsbtRequest/index.tsx
+++ b/src/app/screens/signPsbtRequest/index.tsx
@@ -85,7 +85,7 @@ function SignPsbtRequest() {
 
   const handlePsbtParsing = useCallback(() => {
     try {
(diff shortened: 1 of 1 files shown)


Imamah-Zafar:
we have an exact same PR already open https://github.com/secretkeylabs/xverse-web-extension/pull/459

yknl:
Duplicate

PR #471, opened 2023-06-12
fix: fixes deregister listener bug

victorkirov:
Fixes a reference error for deregistering a listener

# PR Type

- [x] Bugfix

(Optional) Resolved: 
I'm not sure how this will affect the app but we are deregistering a listener that we didn't register in the first place. It could result in a listener lingering in the background and resources not being closed as expected, but I have the feeling users won't really notice it.

# What is the new behavior?
We deregister the tab close listener correctly.


yknl:
Is this PR still needed?

m-aboelenein (approved):

DuskaT021:
this can be merged 👍

m-aboelenein (approved):

PR #479, opened 2023-06-14
fix: don't fire bitcoin and stacks wallet queries when no address

victorkirov:
# PR Type
- [x] Bugfix

When we are creating a new wallet, we don't have a wallet in the store, but the calls for Stacks and Bitcoin wallet data are firing off with an empty address and there is a constant retry in dev tools resulting in endless 404s.

# What is the new behavior?
We only make those calls if there is a bitcoin/stacks wallet address in the store.


DuskaT021:
Tested 👍

m-aboelenein (approved):

PR #481, opened 2023-06-14
fix: fix the confirm password logic

victorkirov:
# PR Type

- [x] Bugfix

When creating a new wallet and you went through the "Backup Wallet" flow, if your password and confirmed password don't match, we don't display the error message as it's wrapped in a try catch instead of in the else.

# What is the new behavior?
If the passwords don't match, we show the error message


m-aboelenein:
@victorkirov needs conflicts resolved

m-aboelenein (approved):

PR #487, opened 2023-06-16
fix: wallet sync non-vanilla objects failure

victorkirov:
# PR Type

- [x] Bugfix
- [ ] Enhancement
- [ ] Build related changes
- [ ] CI related changes

(Optional) Resolved: ENG-2265
If more than one chrome window is open and both windows have the XVerse popup open, then we sync the entire state between them, causing some objects to not deserialise properly. This causes issues when we try to use class obejcts that deserialised into native JS objects. This was evident in the balances shown in the home screen.

# What is the new behavior?
We only sync the required bits of the store, so we don't end up with invalid objects in our store.


m-aboelenein (approved):

DuskaT021:
This is tested and can be merged 👍

m-aboelenein (approved):

PR #495, opened 2023-06-30
feat: add guard for import ledger page

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
We want to prevent the import ledger screen from being open more than once. We do this by closing any open import ledger tabs when a new one opens, similar to onboarding.

Issue Link: #[ENG-2340](https://linear.app/xverseapp/issue/ENG-2340/onboarding-guard-for-the-import-ledger-account-flow-multiple-tabs)

# 🔄 Changes
- Single tab guard was moved to its own component and used in the import ledger route, the onboarding guard, and the reset wallet guard
- When a wallet is reset, we now close all tabs

To be tested:
- Onboarding flow should be retested to ensure single tabs still work as expected
- Import ledger flow needs to be tested for single tabs
- We now close all open tabs when locking or resetting the wallet

# ✅ Review checklist


dhriaznov (approved):
LGTM

yknl (approved):
Tested and working well.

PR #502, opened 2023-07-05
Chore/release workflow

teebszet:
# 🔘 PR Type

- [ ] Bugfix
- [ ] Enhancement
- [x] Build related changes
- [x] CI related changes

# 📜 Background
Previously the release workflow was failing for every release created here: https://github.com/secretkeylabs/xverse-web-extension/actions/workflows/release.yml?page=2

This PR fixes that workflow and the idea is so upload built release artifact to the releases page like here: https://github.com/secretkeylabs/xverse-web-extension/releases/tag/v0.13.0 (needs testing on next release)

This PR also saves the built output of each build workflow run, so that any tester can download the build directly from either the workflow run here: https://github.com/secretkeylabs/xverse-web-extension/actions/runs/5460899929
or as a comment on the PR (needs testing after merge) 

Issue Link: https://linear.app/xverseapp/issue/ENG-2321/fix-release-workflow-for-web-extension
Context Link (if applicable):

# 🔄 Changes

- upload artifact from build workflow and attach as comment to PR
- upload artifact from release workflow and upload to release url
- fix webpack build issues from ubuntu (there was a casing error not detected when building with git checkout on MacOS)

Impact:
- no impact on products
- just easier for testing and release management

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


.github/workflows/release.yml
```diff
@@ -1,17 +1,20 @@
 name: Release
 
 on:
+  workflow_dispatch:
+  # push:
+  #   branches:
+  #     - 'chore/release-workflow'
```
victorkirov, line 7: Please remove the commented out lines

.github/workflows/comment_pr.yml
```diff
@@ -0,0 +1,55 @@
...
+          JOB_PATH: "${{ github.server_url }}/${{ github.repository }}/actions/\
+            runs/${{ env.PREVIOUS_JOB_ID }}"
+          HEAD_SHA: ${{ env.HEAD_SHA }}
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          issue-number: ${{ env.PR_NUMBER }}
+          body: |-
+            build: ${{ env. ARTIFACT_URL }}
+
+            Please note that files only stay for around 90 days
```
victorkirov, line 55: Isn't the retention set to 30 days? 👀

.github/workflows/comment_pr.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Update Comment
+        env:
+          JOB_PATH: "${{ github.server_url }}/${{ github.repository }}/actions/\
+            runs/${{ env.PREVIOUS_JOB_ID }}"
+          HEAD_SHA: ${{ env.HEAD_SHA }}
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          issue-number: ${{ env.PR_NUMBER }}
+          body: |-
+            build: ${{ env. ARTIFACT_URL }}
```
victorkirov, line 53: Not sure if this space will make a difference or not

```suggestion
            build: ${{ env.ARTIFACT_URL }}
```

.github/workflows/comment_pr.yml
```diff
@@ -0,0 +1,55 @@
...
+          HEAD_SHA=$(jq -r '.pull_requests[0].head.sha' \
+            <<< "$WORKFLOW_RUN_EVENT_OBJ")
+
+          echo "Head sha: $HEAD_SHA"
+          echo "HEAD_SHA=$HEAD_SHA" >> "$GITHUB_ENV"
+      - name: Update Comment
+        env:
+          JOB_PATH: "${{ github.server_url }}/${{ github.repository }}/actions/\
+            runs/${{ env.PREVIOUS_JOB_ID }}"
+          HEAD_SHA: ${{ env.HEAD_SHA }}
```
victorkirov, line 48: Where are these env vars used? I don't see them in the docs for `peter-evans/create-or-update-comment@v2` 🤔

.github/workflows/comment_pr.yml
victorkirov, line 1: This workflow is extremely complicated and it always runs after the build workflow. Would it not be simpler to just stick it in that workflow as a job that is dependent on the previous job? Then you can easily reference all the outputs that the `Get Artifact URL & PR Info` step is figuring out?

victorkirov (changes requested):
Added a few comments for consideration.

webpack/utils/build.js
```diff
@@ -8,6 +8,14 @@ var webpack = require('webpack'),
 
 config.mode = 'production';
 
-webpack(config, function (err) {
+webpack(config, function (err, stats) {
   if (err) throw err;
+  console.log(
```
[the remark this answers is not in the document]
teebszet: This was intentional. Previously the web pack build was failing silently, but with this we have output in the workflow logs

src/app/components/alertMessage/index.tsx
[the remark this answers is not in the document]
teebszet: Haha yeah I only found it after I came back to this a few days later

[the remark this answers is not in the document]
teebszet: I think you might be able to set it in git config? Just a hunch, didn't look deeper

.github/workflows/comment_pr.yml
teebszet: Yes! and I was hopeful that would work, but when I was testing with this commit https://github.com/secretkeylabs/xverse-web-extension/pull/502/commits/a5ac31e41a653aacabdcea1fcf6abc6ec3c28d76#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L50-L75, somehow the artifacts REST endpoint didn't return any artifacts while run from the same workflow run.

Although I couldn't find any docs explicitly stating this, this pattern was shown in the docs as an example: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow

webpack/utils/build.js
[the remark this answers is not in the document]
victorkirov: Ok, awesome :)

.github/workflows/comment_pr.yml
victorkirov: 👀

victorkirov: Oh I see, the artifact URL is only available after the upload. I thought maybe `actions/upload-artifact@v3` might have the URL as an output :/

I'll go through this file again to try and figure out  exactly what it's doing 😅

teebszet: > Oh I see, the artifact URL is only available after the upload. I thought maybe actions/upload-artifact@v3 might have the URL as an output :/

there is this issue asking for it lol: https://github.com/actions/upload-artifact/issues/382

teebszet: and this for getting artifacts url: https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28

victorkirov (approved):

PR #510, opened 2023-07-09
Mahmoud/eng 2284 seedphrase vault

m-aboelenein:
# 🔘 PR Type

This PR introduces the usage of the seed vault a class that manages the storage and retrieval of the seed phrase in the web extension

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background

This changes are implemented in response of comments to better enhance the security of the app in regards to users secret keys

# 🔄 Changes
- introduction of the `useSeedVault` hook to handle interactions with the seed vaults methods and initializes the seed vault instance
- created session storage utils to get/add/remove keys
- export one instance of chrome storage to be used across the app
- during create/restore wallet flow the seed phrase is passed to the seed vault for encryption and storage
- changing the encryption algorithm used to encrypt the seed from `AES-CBC` to `AES-GCM`
- changing the password hashing algorithm from `Argon2i` to `Argon2id`
- depending on the `accountsList` to check if a wallet exists or not instead of the `encryptedSecretKey`
- when restoring the user session the password hash and the encrypted seed are fetched from secures storage and used to decrypt the seed phrase
- when the user enters the password it's used to decrypt the seed phrase to verify the password is correct.
- the `seedPhrase` and `encryptedSeed` are removed from global state and are only available in memory during the execution of wallet actions that require the seed phrase

Impact:
- changing how we interact and used the seedPhrase through the context of the app
- added a migration step for all users to use the new vault 

Bugs Resolved:

https://linear.app/xverseapp/issue/ENG-2319/false-incorrect-password-error

# ✅ Review checklist


src/app/utils/encryptionUtils.ts
```diff
@@ -2,24 +2,26 @@ import {
...
-function generateRandomKey(bytesCount: number): string {
+export function generateRandomKey(bytesCount: number): string {
   const randomValues = Array.from(crypto.getRandomValues(new Uint8Array(bytesCount)));
   return randomValues.map((val) => `00${val.toString(16)}`.slice(-2)).join('');
 }
 
-async function generateKeyArgon2(password: string, salt: string): Promise<string> {
+export async function generateKeyArgon2(password: string, salt: string): Promise<string> {
+  console.log("🚀 ~ file: encryptionUtils.ts:15 ~ generateKeyArgon2 ~ salt:", salt)
+  console.log("🚀 ~ file: encryptionUtils.ts:15 ~ generateKeyArgon2 ~ password:", password)
```
teebszet, line 16: remove console log statements

src/app/hooks/useWalletSession.ts
```diff
@@ -12,24 +14,17 @@ const useWalletSession = () => {
...
   };
 
   const shouldLock = async () => {
-    const pHash = await chrome.storage.session.get('pHash');
-
+    const pHash = await getSessionItem(SeedVaultStorageKeys.PASSWORD_HASH);
     if (!pHash) return false;
-
-    const startTime = await getSessionStartTime();
+    const startTime = await await getSessionItem(SESSION_START_TIME_KEY);
```
teebszet, line 27: await await?

src/pages/Popup/index.tsx
```diff
@@ -1,19 +1,22 @@
...
 async function restoreSession() {
-  const { pHash } = await chrome.storage.session.get('pHash');
-  const { walletState } = rootStore.store.getState();
-  if (pHash) {
-    const seed = await decryptMnemonic(walletState.encryptedSeed, pHash);
+  const pHash = await getSessionItem(SeedVaultStorageKeys.PASSWORD_HASH);
+  const encryptedSeed = await ChromeStorage.getItem(SeedVaultStorageKeys.ENCRYPTED_KEY);
+  if (pHash && encryptedSeed) {
+    const seed = await decryptSeedPhrase(encryptedSeed as string, pHash);
     rootStore.store.dispatch(setWalletSeedPhraseAction(seed));
```
teebszet, line 20: do we still need to store the decrypted seed phrase into the store here?

src/app/components/guards/auth.tsx
```diff
@@ -1,22 +1,48 @@
...
-  return <>{children}</>;
+  useEffect(() => {
+    restoreSession();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  if (!authTested) {
+    return null;
+  }
+  return children;
```
teebszet, line 45: the fragment was required to fix eslint errors, as children can be an array of ReactNodes

src/app/hooks/useSecretKey.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const clearVaultStorage = async () => {
+    await ChromeStorage.removeItem(SeedVaultStorageKeys.ENCRYPTED_KEY);
+    await ChromeStorage.removeItem(SeedVaultStorageKeys.PASSWORD_SALT);
+    await removeSessionItem(SeedVaultStorageKeys.PASSWORD_HASH);
+  };
+
+  return {
+    ...vault,
+    initSeedVault: vault.init,
```
teebszet, line 48: why do you need this returned twice?

src/app/hooks/useSecretKey.ts
teebszet, line 1: feels like the useSecretKey name is overloaded. why not useSeedVault.ts ?

src/app/hooks/useSignatureRequest.ts
teebszet, line 1: nit: should run prettier on this file

teebszet (changes requested):

src/app/components/accountHeader/index.tsx
```diff
@@ -154,6 +154,11 @@ function AccountHeaderComponent({
     setShowOptionsDialog(false);
   };
 
+  const handleLockWallet = async () => {
+    await lockWallet();
+    navigate('/login');
```
victorkirov, line 159: Won't this be handled by the guard?

src/app/components/guards/auth.tsx
```diff
@@ -1,22 +1,48 @@
...
+  const restoreSession = async () => {
+    if (encryptedSeed) {
+      navigate('/login');
+      setAuthTested(true);
+      return;
+    }
+    const hasSeedPhrase = await hasSeed();
+    if (!hasSeedPhrase || !masterPubKey) {
+      navigate('/landing');
+      setAuthTested(true);
```
victorkirov, line 25: I don't think this does anything. Once you navigate, the local state is lost.

src/app/components/guards/auth.tsx
victorkirov: Maybe we should add a comment?
```suggestion
  // fragment is required here because without it, the router thinks there could be more than 1 child node
  // eslint-disable-next-line react/jsx-no-useless-fragment
  return children;
```

src/app/components/guards/onboarding/index.tsx
```diff
@@ -23,18 +21,29 @@ function OnboardingGuard({ children }: WalletExistsGuardProps): React.ReactEleme
...
     () => ({
       disableWalletExistsGuard: () => setWalletExistsGuardEnabled(false),
     }),
     [],
   );
+  useEffect(() => {
+    (async () => {
+      try {
+        const seed = await getSeed();
+        setWalletSeedPhrase(seed);
```
victorkirov, line 36: Aren't we unnecessarily putting the seed in memory here? Maybe this should be the `hasSeed` method and we just store a boolean since we aren't actually using the seed?

src/app/hooks/useSignatureRequest.ts
```diff
@@ -82,18 +85,17 @@ export function useSignMessage(messageType: SignatureMessageType) {
...
+  const {getSeed} = useSecretKey();
+  return useCallback(async () => {
+    const seedPhrase = await getSeed();
+   return signBip322Message({
+    accounts: accountsList,
+    message,
+    signatureAddress: address,
+    seedPhrase,
+    network: network.type,
+  })}, []);
```
victorkirov, line 98: This should probably have dependencies listed.

src/app/screens/sendBtc/index.tsx
```diff
@@ -154,7 +157,7 @@ function SendBtcScreen() {
     ];
     setRecipient(recipients);
     if (validateFields(address, amountToSend)) {
-      mutate({ recipients });
+       mutate({ recipients, seedPhrase });
```
victorkirov, line 160: Space at start 👀 You should probably enable format on save as per the Vscode page in Notion

src/app/utils/encryptionUtils.ts
```diff
@@ -25,39 +38,37 @@ async function generateKeyArgon2(password: string, salt: string): Promise<string
...
 
 export async function generatePasswordHash(password: string) {
-  const existingSalt = getSalt();
-  if (existingSalt) {
-    const argonHash = await generateKeyArgon2(password, existingSalt);
-    return {
-      salt: existingSalt,
-      hash: argonHash,
-    };
+  let salt = await ChromeStorage.getItem(SeedVaultStorageKeys.PASSWORD_SALT);
```
victorkirov, line 41: It's a bit of a strange pattern that we are using the `SeedVaultStorageKeys` directly in the code instead of via the seed vault. Is there way to do something like `seedVault.getSalt()`?

src/app/utils/storage.ts
```diff
@@ -30,7 +30,7 @@ class ChromeStorage {
     });
   }
 
-  getItem(key: string) {
+  getItem(key: string): Promise<any> {
```
victorkirov, line 33: Might be nice to make this generic but default to any:
```suggestion
  getItem<T = any>(key: string): Promise<T> {
```

src/app/hooks/useWalletReducer.ts
```diff
@@ -102,34 +105,43 @@ const useWalletReducer = () => {
...
+      return decrypted;
+    });
+  };
+
+  const unlockWallet = async (password: string) => {
+    if (encryptedSeed && encryptedSeed.length > 0) {
+      await migrateLegacySeedStorage(password);
+      return;
+    }
+    const decrypted = await unlockVault(password);
```
victorkirov, line 124: Won't this result in the unlock flow not completing on migration? Should it not continue after the migration and not return yet since it doesn't perform the functions after line 124?

```suggestion
    let decrypted: string
    if (encryptedSeed && encryptedSeed.length > 0) {
      decrypted = await migrateLegacySeedStorage(password);
    } else {
      decrypted = await unlockVault(password);
    }

// ... continue with funciton
```

src/app/hooks/useWalletReducer.ts
```diff
@@ -102,34 +105,43 @@ const useWalletReducer = () => {
...
       dispatch(fetchAccountAction(accountsList[0], accountsList));
       dispatch(getActiveAccountsAction(accountsList));
     } finally {
-      chrome.storage.session.set({
-        pHash: pHash.hash,
-      });
       setSessionStartTime();
     }
-    dispatch(unlockWalletAction(decrypted));
     return decrypted;
```
victorkirov, line 133: Don't think we should return the seed phrase here. Opens it up to accidental in-memory storage. Rather only ever expose the seed phrase form the vault.

src/app/hooks/useWalletReducer.ts
```diff
@@ -102,34 +105,43 @@ const useWalletReducer = () => {
...
 
-  const resetWallet = () => {
+  const resetWallet = async () => {
     dispatch(resetWalletAction());
-    chrome.storage.local.clear();
-    chrome.storage.session.clear();
+    await chrome.storage.local.clear();
+    await chrome.storage.session.clear();
     localStorage.clear();
     clearSessionTime();
```
victorkirov, line 146: No need to do the async calls synchronously

```suggestion
localStorage.clear();
await Promise.all([
  clearSessionTime(),
  chrome.storage.local.clear(),
  chrome.storage.session.clear()
])
```

src/app/screens/createPassword/index.tsx
```diff
@@ -49,20 +46,19 @@ function CreatePassword(): JSX.Element {
...
   useEffect(() => {
-    if (!seedPhrase) {
-      navigate('/backup');
-    }
-  }, [seedPhrase]);
+    (async () => {
+      const seed = await getSeed();
+      setSeedPhrase(seed);
+    })();
+  }, []);
```
teebszet, line 61: why do we use the `useState` pattern in this screen, and not call `await getSeed()` within `handleConfirmPassword` handler, as in other screens?

src/app/screens/backupWallet/index.tsx
```diff
@@ -58,14 +57,14 @@ const TransparentButtonContainer = styled.div((props) => ({
...
 
   useEffect(() => {
-    if (!seedPhrase) {
+    (async () => {
+      await initSeedVault('');
       const newSeedPhrase = generateMnemonic();
-      storeSeedPhrase(newSeedPhrase);
-    }
+      await storeSeed(newSeedPhrase);
+    })();
```
teebszet, line 67: the behaviour changed here. need to check why 
```
if (!seedPhrase) {
}
```

is no longer required

src/app/hooks/useWalletSession.ts
```diff
@@ -12,24 +14,17 @@ const useWalletSession = () => {
...
-  const getSessionStartTime = async () => {
-    const { sessionStartTime } = await chrome.storage.session.get(SESSION_START_TIME_KEY);
-    return sessionStartTime;
+    await removeSessionItem(SESSION_START_TIME_KEY);
   };
 
   const shouldLock = async () => {
-    const pHash = await chrome.storage.session.get('pHash');
-
+    const pHash = await getSessionItem(SeedVaultStorageKeys.PASSWORD_HASH);
```
teebszet, line 25: why isn't this in the SeedVault class?

package-lock.json
fedeerbes, line 1: watch out with this diff folks. i imagine this what happened to me on mobile as well https://secretkeylabs.slack.com/archives/C05QCJM9D1N/p1694524281019429 if not ignore this comment =)

teebszet: https://stackoverflow.com/questions/75224019/resolved-and-integrity-go-missing-in-package-lock-json if you are running into this issue you can try with the suggested. i solved my issue was again by solving merge conflicts and then reinstalling deps. next time i'll try this SO suggestion

src/app/components/accountHeader/index.tsx
m-aboelenein: since the seed phrase and encrypted seed are no longer considered global state the `Authguard` won't re-render and trigger the effect when locking the vault the only alternative here being adding a listener on storage and storing a flag on the wallet store

src/app/components/guards/auth.tsx
m-aboelenein: resolved

src/app/components/guards/auth.tsx
```diff
@@ -1,22 +1,48 @@
...
-  const { encryptedSeed, seedPhrase } = useWalletSelector();
-  const hydrated = useHasStateRehydrated();
+interface AuthGuardProps {
+  children: ReactNode;
+}
 
-  if (hydrated && encryptedSeed && !seedPhrase) {
-    return <Navigate to="/login" />;
-  }
+function AuthGuard({ children }: AuthGuardProps) {
```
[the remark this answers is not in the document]
m-aboelenein: resolved

src/app/utils/storage.ts
```diff
@@ -19,7 +19,7 @@ class ChromeStorage {
     return runtimeMap.get(this).lastError;
   }
 
-  setItem(key: string, item: any) {
+  setItem(key: string, item: any): Promise<any> {
```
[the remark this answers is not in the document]
m-aboelenein: nope actually the type from the docs here is void `https://developer.chrome.com/docs/extensions/reference/storage/#type-StorageArea`

src/app/utils/sessionStorageUtils.ts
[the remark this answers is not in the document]
m-aboelenein: implemented implemented the second approach

src/app/utils/encryptionUtils.ts
```diff
@@ -25,39 +38,37 @@ async function generateKeyArgon2(password: string, salt: string): Promise<string
...
+  const unMigratedSalt = localStorage.getItem('salt');
+  if (!salt && !unMigratedSalt) {
+    salt = generateRandomKey(16);
+  } else {
+    salt = unMigratedSalt || salt;
```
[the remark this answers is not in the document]
m-aboelenein: it's cleared after completing `migrateLegacySeedStorage`

src/app/utils/encryptionUtils.ts
```diff
@@ -25,39 +38,37 @@ async function generateKeyArgon2(password: string, salt: string): Promise<string
...
   }
-  const newSalt = generateRandomKey(16);
-  saveSalt(newSalt);
-  const argonHash = await generateKeyArgon2(password, newSalt);
+  const argonHash = await generateKeyArgon2i(password, salt);
```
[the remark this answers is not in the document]
m-aboelenein: the seed vault is using argon2id while legacy is argon2i

src/app/hooks/useSecretKey.ts
```diff
@@ -0,0 +1,52 @@
+import { useMemo } from 'react';
+import SeedVault, {
+  CryptoUtilsAdapter,
+  SeedVaultStorageKeys,
+  StorageAdapter,
+} from '@secretkeylabs/xverse-core/seedVault';
+import { generateRandomKey } from '@secretkeylabs/xverse-core/encryption';
+import { getSessionItem, removeSessionItem, setSessionItem } from '@utils/sessionStorageUtils';
+import * as cryptoUtils from '@utils/encryptionUtils';
```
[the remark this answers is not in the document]
m-aboelenein: resolved

src/app/utils/storage.ts
```diff
@@ -52,4 +52,4 @@ class ChromeStorage {
     });
   }
 }
-export default ChromeStorage;
+export default new ChromeStorage(chrome.storage.local, chrome.runtime);
```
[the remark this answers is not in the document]
m-aboelenein: resolved

src/app/hooks/useSecretKey.ts
```diff
@@ -0,0 +1,52 @@
...
+        commonStorageAdapter,
+      }),
+    [],
+  );
+
+  const clearVaultStorage = async () => {
+    await ChromeStorage.removeItem(SeedVaultStorageKeys.ENCRYPTED_KEY);
+    await ChromeStorage.removeItem(SeedVaultStorageKeys.PASSWORD_SALT);
+    await removeSessionItem(SeedVaultStorageKeys.PASSWORD_HASH);
+  };
```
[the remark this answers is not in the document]
m-aboelenein: i updated the core-lip pr 👍

src/app/hooks/useSecretKey.ts
m-aboelenein: initially to make it more readable because init is a little vague removed this in favor of aliasing the function where it's being used

src/app/hooks/useSecretKey.ts
m-aboelenein: agreee

src/app/hooks/useSignatureRequest.ts
m-aboelenein: resolved

src/app/hooks/useSignatureRequest.ts
m-aboelenein: no need since no dependencies change for the same request

src/app/screens/sendBtc/index.tsx
m-aboelenein: done

src/app/utils/encryptionUtils.ts
m-aboelenein: refactored the function so it will only deal with legacy so no need to support getting individual stored items anymore

src/app/utils/storage.ts
m-aboelenein: nice

src/app/components/guards/auth.tsx
m-aboelenein: resolved however i found if we don't return after navigating the effect continues to check for the remaining conditions and ends up defaulting to the login screen

src/app/hooks/useWalletReducer.ts
m-aboelenein: it was decided to not load the accounts when undergoing the initial migration

src/app/hooks/useWalletReducer.ts
m-aboelenein: updated

src/app/hooks/useWalletReducer.ts
m-aboelenein: resolved

src/app/screens/createPassword/index.tsx
m-aboelenein: true, resolved

src/app/hooks/useWalletSession.ts
m-aboelenein: moved the behavior to the seedvault as `isVaultUnlocked`

src/app/screens/backupWallet/index.tsx
m-aboelenein: it was no longer necessary since the seedphrase can no longer be synced between tabs and we removed forward/backward capabilities

abdulhaseeb4239 (approved):

teebszet:
thanks for resolving those 🕺 I leave the rest to @victorkirov

src/app/components/accountHeader/index.tsx
victorkirov: Oh, that's not ideal. We want to be reactive to changes in the seed vault. I think we should make the following changes:
- In the seed vault repo in core, we add an event publisher to the seedvault class that you can listen on and whenever a function that causes a change in the seedvault is fired, we notify the listeners.
- In the extension/mobile app, we create a singleton of the seedvault and expose it via Context API. We can then change the useSeedVault hook to wrap useContext and we provide the singleton seedvault as the value in the provider.
- The context then listens to changes to the seedvault and regenerates the singleton to a new instance of the seedvault and passes it down as the new value. retriggering anything that's using the seedvault hook.

What do you think?

src/app/components/guards/auth.tsx
victorkirov: That makes sense. I think returning is the better approach.

src/app/screens/backupWallet/index.tsx
victorkirov: This seems rally dangerous 👀 If we introduce a bug at some point and accidentally call `await storeSeed(newSeedPhrase);` the seedvault will overwrite the old one without issue because it doesn't check if there isn't already a stored seed. I think we need ot throw an error if we try to set a seed when one already exists and then rethink this flow a little.

m-aboelenein:
@DuskaT021 @victorkirov this is ready for Review and QA

m-aboelenein:
> @DuskaT021 @victorkirov this is ready for Review and QA

we can begin QA first

package-lock.json
teebszet, line 1: @m-aboelenein can we revert the changes to this file, and only update what was changed in package.json

package.json
```diff
@@ -7,11 +7,11 @@
     "@ledgerhq/hw-transport-webusb": "^6.27.13",
     "@phosphor-icons/react": "^2.0.10",
     "@react-spring/web": "^9.6.1",
-    "@secretkeylabs/xverse-core": "1.9.1",
+    "@secretkeylabs/xverse-core": "^1.9.1-99ed768",
```
teebszet, line 10: reminder: need to change this version

src/manifest.json
```diff
@@ -21,7 +21,7 @@
   "host_permissions": ["*://*/*"],
   "permissions": ["storage"],
   "content_security_policy": {
-    "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; object-src 'self'; frame-src 'none'; frame-ancestors 'none';"
+    "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; object-src 'none'; default-src 'none'; connect-src *; style-src 'unsafe-inline'; frame-src 'none'; frame-ancestors 'none'; img-src 'self' data: https:;  media-src 'self' data: https:; "
```
teebszet, line 24: could we have a brief explanation on this?

src/pages/Popup/index.tsx
```diff
@@ -9,27 +6,14 @@ import './index.css';
...
   persistQueryClient({
     queryClient,
     persister: offlineStorage,
     buster: VERSION,
   });
   const container = document.getElementById('app');
 
-  // We need to do this after load in order for the extension popup to have the correct height on first load
-  container!.style.maxHeight = '100vh';
-
```
teebszet, line 32: @victorkirov do we not need this anymore?

victorkirov: Nope, this was replaced by another mechanism. See the useEffect here:
https://github.com/secretkeylabs/xverse-web-extension/blob/f0b289c725a8b447fc6918b4c2ac62537658a664/src/app/components/screenContainer/index.tsx

src/app/screens/backupWalletSteps/index.tsx
```diff
@@ -39,19 +36,22 @@ export default function BackupWalletSteps(): JSX.Element {
...
-    }
-  }, [seedPhrase]);
+    (async () => {
+      try {
+        const seed = await getSeed();
+        setSeedPhrase(seed);
+      } catch (e) {
+        navigate('/backup');
+      }
+    })();
```
teebszet, line 53: no destroyer here, as in other usages?

teebszet: also, is the changed behaviour intentional? do we want to use this?

```
if (!hasSeed()) { navigate('/backup')} 
```

src/common/utils/messageHandlers.ts
```diff
@@ -1,42 +1,19 @@
...
 
-let inMemoryKey = '';
-
 async function internalBackgroundMessageHandler(
   message: BackgroundMessages,
   sender: chrome.runtime.MessageSender,
   sendResponse: (response?: any) => void,
 ) {
   if (!validateMessagesAreFromExtension(sender)) {
+    sendResponse();
```
teebszet, line 14: why did this logic change?

it looks like the if block is redundant now

src/app/screens/settings/backupWallet/index.tsx
```diff
@@ -89,7 +99,7 @@ function BackupWalletScreen() {
         )}
         <SeedphraseContainer>
           {showSeed && (
-            <SeedCheck showButton={false} seedPhrase={seedPhrase} onContinue={goToSettingScreen} />
+            <SeedCheck showButton={false} seedPhrase={seed} onContinue={goToSettingScreen} />
```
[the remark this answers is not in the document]
m-aboelenein: in this case it should be ok the seedphrase is displayed in plain text here and once the component is removed from the dom tree it should be cleaned up

src/manifest.json
m-aboelenein: this was added in response to the incorrect password issue that happened while testing and for some user when updating the extension for more context on the issue itself please check the Gotcha section form here [Gotchas](https://www.stackhawk.com/blog/react-content-security-policy-guide-what-it-is-and-how-to-enable-it/)

src/common/utils/messageHandlers.ts
m-aboelenein: we were utilizing the background worker for storing the `seedphrase` in memory we don't do that anymore with the `SeedVault` , sendResponse is only there to indicate that the handler was terminated
 ```
 export enum InternalMethods {
  RequestDerivedStxAccounts = 'RequestDerivedStxAccounts',
  ShareInMemoryKeyToBackground = 'ShareInMemoryKeyToBackground',
  RequestInMemoryKeys = 'RequestInMemoryKeys',
  RemoveInMemoryKeys = 'RemoveInMemoryKeys',
  OriginatingTabClosed = 'OriginatingTabClosed',
}
```

src/manifest.json
teebszet: nice one!

src/app/screens/settings/backupWallet/index.tsx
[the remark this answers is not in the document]
teebszet: I think the ref I linked is saying that there is no explicit rule that the memoized values will be removed. it doesn't mention any clean up when the component unmounts, and it seems like a very inexpensive operation anyway, so not sure why it needs to be memoized

package.json
m-aboelenein: the seed vault pr is not merged yet

teebszet (dismissed):

teebszet (approved):

m-aboelenein:
new changes affect: 
- backup wallet screen settings
- request from dapp then closing the dapp tab

teebszet (approved):

PR #538, opened 2023-07-28
Fix regression issues v0.14.0-rc.1

dhriaznov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
There is a number of issues found after regression for the v0.14.0-rc.1 version

Issue Link: https://linear.app/xverseapp/issue/ENG-2473/regression-issues-v0140-rc1
Context Link (if applicable):

# 🔄 Changes
- Fixed "Verify addresses" popup width for ledger accounts on the full screen gallery view
- Fixed "NUMBER OF COLLECTIBLES" after activation on the gallery tab in extension view (it showed 0 even after fetching the items)
- Fixed screen crash when changing account on /send-brc20 screen

# 🖼 Screenshot / 📹 Video
Monosnap Xverse Wallet 2023-07-28 13-06-53

# ✅ Review checklist

### DIFF
--- a/src/app/screens/nftDashboard/receiveNft/index.tsx
+++ b/src/app/screens/nftDashboard/receiveNft/index.tsx
@@ -11,13 +11,6 @@ import { useState } from 'react';
 import { isLedgerAccount } from '@utils/helper';
 import ReceiveCardComponent from '../../../components/receiveCardComponent';
 
-interface Props {
-  visible: boolean;
-  onClose: () => void;
-  setOrdinalReceiveAlert: () => void;
-  isGalleryOpen: boolean;
-}
-
 const ColumnContainer = styled.div((props) => ({
   display: 'flex',
   flexDirection: 'column',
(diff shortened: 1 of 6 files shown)


src/app/screens/nftDashboard/receiveNft/index.tsx
```diff
@@ -133,15 +137,15 @@ function ReceiveNftModal({ visible, onClose, isGalleryOpen, setOrdinalReceiveAle
...
+    <VerifyButtonsContainer>
       <RowContainer>
         <Text>{t('RECEIVE_NFT')}</Text>
         <ButtonImage onClick={onReceiveModalClose}>
           <img src={Cross} alt="cross" />
         </ButtonImage>
       </RowContainer>
       {isReceivingAddressesVisible ? receiveContent : verifyOrViewAddresses}
-    </>
+    </VerifyButtonsContainer>
```
teebszet, line 148: VerifyButtonsContainer name is misleading, as it sometimes contains receiveContent and no verify buttons

why not just set a minWidth on VerifyButtonContainer, which existed already? then you don't need an extra div

src/app/screens/sendBrc20/index.tsx
```diff
@@ -96,12 +96,12 @@ function SendBrc20Screen() {
...
   };
 
-  const checkIfEnableButton = () => {
+  const checkIfEnableButton = useMemo(() => {
     if (amountToSend !== '' || amountToSend <= fungibleToken.balance) {
       return true;
     }
     return false;
-  };
+  }, [amountToSend, fungibleToken]);
```
teebszet, line 104: this is not really an expensive calculation worth memoizing. since amountToSend is component state, it will already trigger a re-render, so you could simplify to:

`const isSendEnabled = !!(amountToSend !== '' || amountToSend <= fungibleToken.balance);`

teebszet: on closer look, the existing logic is broken. this should have been

`const isSendEnabled = amountToSend !== '' && amountToSend <= fungibleToken.balance`

but also the button doesn't get disabled, it just gets an opacity change. should fix by using the `disabled` prop on ActionButton

dhriaznov:
I updated this PR @teebszet, could you please take a look once again?

teebszet (approved):
Lgtm

PR #550, opened 2023-08-03
Add an override switch for sponsoring transaction in swap screen

Imamah-Zafar:
# 🔘 PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
All ALEX swaps transactions currently get broadcasted to the stacks-transaction-sponsor service via the react hook
In the case that the sponsor service is active, but returns an error that we know is coming from the sponsor service, we want to offer feedback to the user to try again, but without attempting to use the sponsor service.

Issue Link: [#[ENG-2527]](https://linear.app/xverseapp/issue/ENG-2527/add-an-override-switch-for-sponsoring-transaction-in-swap-screen)

# 🔄 Changes
The hook `useSponsoredTransaction` has been modified so it accepts a function argument `isSponsorOptionSelected` which is set using the toggle shown under the details option in the swap screen. The `isSponsored` value is set true when the `isServiceRunning` & `isSponsorOptionSelected` both are true. The `isServiceRunning` value depends on the sponsor service api is running. 
The toggle is shown in the swap screen depending if `isServiceRunning` is true. 
In order to give feedback to user to try again the transaction if the error is due to sponsor service, a `swapErrorList` is introduced which will have all the known errors and if the error returned while broadcasting the transaction is present in the error we show users the information note on confirmation screen

# 🖼 Screenshot / 📹 Video

**When sponsor service is up**

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/14722d18-2c58-4737-ae67-49d0bb2310c1

**When sponsor service is down**

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/3b00b933-a139-4dad-ab1c-d1195969fb3e

# ✅ Review checklist

### DIFF
--- a/src/app/screens/swap/swapInfoBlock/index.tsx
+++ b/src/app/screens/swap/swapInfoBlock/index.tsx
@@ -1,11 +1,22 @@
 import { UseSwap } from '@screens/swap/useSwap';
 import { useState } from 'react';
-import styled from 'styled-components';
+import styled, { useTheme } from 'styled-components';
 import { useTranslation } from 'react-i18next';
 import SlippageEditIcon from '@assets/img/swap/slippageEdit.svg';
 import ChevronIcon from '@assets/img/swap/chevron.svg';
 import BottomModal from '@components/bottomModal';
 import { SlippageModalContent } from '@screens/swap/slippageModal';
+import Switch from 'react-switch';
+
+const CustomSwitch = styled(Switch)`
+  .react-switch-handle {
+    background-color: ${({ checked }) =>
+      checked ? '#FFFFFF' : 'rgba(255, 255, 255, 0.2)'} !important;
+    border: ${({ checked }) => (checked ? '' : '4px solid rgba(255, 255, 255, 0.2)')} !important;
+    height: 16px !important;
+    width: 16px !important;
+  }
+`;
 
 const PoweredByAlexText = styled.span((props) => ({
   ...props.theme.body_xs,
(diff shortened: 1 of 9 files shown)


src/app/screens/swap/swapInfoBlock/index.tsx
```diff
@@ -84,6 +106,23 @@ export function SwapInfoBlock({ swap }: { swap: UseSwap }) {
...
+                <ToggleContainer>
+                  <CustomSwitch
+                    onColor={theme.colors.purple_main}
+                    offColor={theme.colors.background.elevation3}
+                    onChange={toggleFunction}
+                    checked={swap.isSponsored}
+                    uncheckedIcon={false}
+                    checkedIcon={false}
+                    height={19}
+                    width={36}
```
teebszet, line 121: how come there is height/width here as well as in CSS on L16,17?

Imamah-Zafar: On L16,17 the height and width is set for the react-switch-handle which is the circle within the switch. Here the main switch body dimensions are set. We need to alter the two since according to the design we are not using the default dimensions

teebszet: ok, let's clean this up when we build out the UI revamp

teebszet (approved):

teebszet:
tested, lgtm. will need a retest after https://linear.app/xverseapp/issue/ENG-2548/integrate-with-alexs-new-broadcastsponsoredtx-function-and-healthz

PR #556, opened 2023-08-08
Fix ui glitch on the login screen

Imamah-Zafar:
# 🔘 PR Type

- [x] Bugfix
- [ ] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
If the button entered into loading state then the entire screen would flicker. 

Issue Link: #[ENG-2571](https://linear.app/xverseapp/issue/ENG-2571/ui-glitch-on-the-login-screen-not-an-issue-on-prod-0141)

# 🔄 Changes
The transition css styling has been removed since this was the code that caused flickering on different screens

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/230e7a6f-1d13-4bc3-b1a2-ac68bbda0eb3

# ✅ Review checklist

### DIFF
--- a/src/app/components/button/index.tsx
+++ b/src/app/components/button/index.tsx
@@ -17,16 +17,6 @@ const Button = styled.button<ButtonProps>((props) => ({
     : props.theme.colors.action.classic,
   width: '100%',
   height: 44,
-  transition: 'all 0.1s ease',
-  ...(props.disabled
-    ? {
-        cursor: 'not-allowed',
-        opacity: 0.4,
-      }
-    : {
-        ':hover': { opacity: 0.8 },
-        ':active': { opacity: 0.6 },
-      }),
 }));
 
 const TransparentButton = styled(Button)((props) => ({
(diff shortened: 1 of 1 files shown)


teebszet:
@Imamah-Zafar the disabled CSS styling needs to be there, otherwise a button will not appear to be disabled when you hover or click on it. 

based on your changes, I guess the button using props for styles was causing some extra re-rendering on the page. so I added those styles back using :disabled pseudo class instead of the disabled prop, which seems to have solved it

Imamah-Zafar:
> @Imamah-Zafar the disabled CSS styling needs to be there, otherwise a button will not appear to be disabled when you hover or click on it.
> 
> based on your changes, I guess the button using props for styles was causing some extra re-rendering on the page. so I added those styles back using :disabled pseudo class instead of the disabled prop, which seems to have solved it.
> 
> please review and test

i see . Tested and this works now with button disabled 👍

PR #558, opened 2023-08-09
fix: Don't use ordinals as inputs when recovering ordinals

victorkirov:
# 🔘 PR Type

- [X] Bugfix

NOTE: We need to add the new version of core once the Xord Ordinal UTXO detection PR is merged before this is fully functional

# 📜 Background
When recovering ordinals, we would sometimes spend other ordinals for the transaction. This was because the confirmation screen would regenerate the signed txn hex if the fee was changed, but it wouldn't use the full list of ordinal UTXOs when doing so.

# 🔄 Changes
- Added the ordinal UTXO fetch in the confirm txn component so that we don't spend other ordinals
- Fixed an issue where clicking on `transfer` for an ordinal recovery, all the ordinals on the recover screen would show a loader

# ✅ Review checklist

Core pr: https://github.com/secretkeylabs/xverse-core/pull/203

### DIFF
--- a/src/app/components/confirmBtcTransactionComponent/index.tsx
+++ b/src/app/components/confirmBtcTransactionComponent/index.tsx
@@ -1,33 +1,34 @@
-import TopRow from '@components/topRow';
-import { useTranslation } from 'react-i18next';
-import styled from 'styled-components';
-import { ReactNode, useEffect, useState } from 'react';
-import BigNumber from 'bignumber.js';
-import ActionButton from '@components/button';
-import AssetIcon from '@assets/img/transactions/Assets.svg';
 import SettingIcon from '@assets/img/dashboard/faders_horizontal.svg';
+import AssetIcon from '@assets/img/transactions/Assets.svg';
+import ActionButton from '@components/button';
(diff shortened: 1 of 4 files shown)


src/app/components/confirmBtcTransactionComponent/index.tsx
```diff
@@ -202,24 +203,28 @@ function ConfirmBtcTransactionComponent({
...
+
+      const newSignedTx = await signOrdinalSendTransaction(
         recipients[0]?.address,
         ordinalTxUtxo!,
         btcAddress,
         Number(selectedAccount?.id),
         seedPhrase,
         network.type,
-        [ordinalTxUtxo!],
+        ordinalsUtxos,
```
victorkirov, line 224: This was the main cause

m-aboelenein (approved):

DuskaT021:
This can be merged but needs the xverse-core 1.6.1 @teebszet

teebszet:
Yup Ken already created that one

DuskaT021:
tested it can be merged

PR #568, opened 2023-08-16
Filter out dead pending transactions

Imamah-Zafar:
# 🔘 PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
ALEX's sponsor service tries to broadcast replace by fee transactions and that probably causes an issue on the Stacks API. Creating dead pending transactions which results in user's not being able to use sponsored option for alex swap as it always indicates that there is a transaction pending

Issue Link: #[i[ENG-2589](https://linear.app/xverseapp/issue/ENG-2589/filter-out-dead-pending-transactions)]
Context Link (if applicable):

# 🔄 Changes
To filter out dead pending transactions, we first retrieve the account's current nonce. 

To obtain the account's current nonce, relying solely on pending transactions is insufficient. This is because, in the case of a dead pending sponsor transaction, the user might have initiated other transactions (both pending and confirmed). These transactions would have nonces greater than that of the dead transaction. Therefore, we retrieve the nonce of the confirmed transactions using the state from redux. If transactions made after the dead pending transaction have been confirmed, then the current account nonce would be one value greater than the nonce of the latest confirmed transaction.

Once we have obtained the account's current nonce, we check if it's greater than the pending transaction nonce. If yes we ignore the pending transactions. If not then the pending transactions are checked 

# 🖼 Screenshot / 📹 Video

While opening up PR, I tested with address `SP20B8SK3STWVAP8QPTQH5FSP6YTJBY3M6CNS6BK0` which had a dead pending swap transaction broadcasted 21 hours ago. I had made a few transactions after the dead pending one and all of them were confirmed.  The video below shows how before the fix, the sponsor toggle was off as it showed that there are pending transactions. The after video depicts how the sis enabledion is enables and the dead pending transaction is ignored
**Before**

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/a610c79f-d70f-4ea4-953c-1708bde4208d

**After**

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/6a62b33f-a871-48e2-b086-5a68d3ba39e5

# ✅ Review checklist

### DIFF
--- a/src/app/screens/swap/useAlexSponsoredTransaction.ts
+++ b/src/app/screens/swap/useAlexSponsoredTransaction.ts
@@ -1,8 +1,14 @@
 import { useEffect, useRef, useState } from 'react';
 import { useQuery } from '@tanstack/react-query';
-import { StacksTransaction } from '@secretkeylabs/xverse-core';
+import {
+  getConfirmedTransactions,
+  StacksTransaction,
+  StxTransactionListData,
+} from '@secretkeylabs/xverse-core';
 import { AlexSDK } from 'alex-sdk';
 import useStxPendingTxData from '@hooks/queries/useStxPendingTxData';
+import useWalletSelector from '@hooks/useWalletSelector';
+import useNetworkSelector from '@hooks/useNetwork';
 
 const useAlexSponsorSwapEnabledQuery = (alexSDK: AlexSDK) =>
   useQuery({
@@ -13,7 +19,11 @@ const useAlexSponsorSwapEnabledQuery = (alexSDK: AlexSDK) =>
 export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) => {
   const alexSDK = useRef(new AlexSDK()).current;
   const [isServiceRunning, setIsServiceRunning] = useState(false);
+  const [hasPendingTransactions, setHasPendingTransactions] = useState(false);
   const { error, data: isEnabled, isLoading } = useAlexSponsorSwapEnabledQuery(alexSDK);
+  const { stxAddress } = useWalletSelector();
+  const selectedNetwork = useNetworkSelector();
+  const { data: stxPendingTxData } = useStxPendingTxData();
 
   useEffect(() => {
     if (!isLoading && !error) {
@@ -24,17 +34,35 @@ export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) =
   const sponsorTransaction = async (signed: StacksTransaction) =>
     alexSDK.broadcastSponsoredTx(signed.serialize().toString('hex'));
 
-  const { data: stxPendingTxData } = useStxPendingTxData();
-  const currentNonce =
-    (stxPendingTxData?.pendingTransactions ?? []).reduce(
-      (maxNonce, transaction) => Math.max(maxNonce, transaction?.nonce ?? 0),
-      0,
-    ) + 1;
+  const getAccountCurrentNonce = async () => {
+    let nonce = 0;
+    const confirmedTransactions: StxTransactionListData = await getConfirmedTransactions({
+      stxAddress,
+      network: selectedNetwork,
+    });
 
-  const filteredStxPendingTxData = stxPendingTxData?.pendingTransactions.filter(
-    (transaction) => currentNonce - transaction?.nonce === 1,
-  );
-  const hasPendingTransactions = filteredStxPendingTxData?.length > 0;
+    if (confirmedTransactions && confirmedTransactions.transactionsList.length > 0) {
+      nonce = confirmedTransactions.transactionsList[0].nonce + 1;
+    }
+    const currentPendingNonce =
+      (stxPendingTxData?.pendingTransactions ?? []).reduce(
+        (maxNonce, transaction) => Math.max(maxNonce, transaction?.nonce ?? 0),
+        0,
+      ) + 1;
+
+    return Math.max(currentPendingNonce, nonce);
+  };
+
+  useEffect(() => {
+    (async () => {
+      const currentNonce = await getAccountCurrentNonce();
+      const filteredStxPendingTxData = stxPendingTxData?.pendingTransactions.filter(
+        (transaction) => currentNonce - transaction?.nonce === 1,
+      );
+
+      setHasPendingTransactions(filteredStxPendingTxData?.length > 0);
+    })();
+  }, [stxPendingTxData]);
 
   return {
     isSponsored: userOverrideSponsorValue && isServiceRunning && !hasPendingTransactions,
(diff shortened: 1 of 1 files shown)


src/app/screens/swap/useAlexSponsoredTransaction.ts
```diff
@@ -24,8 +34,35 @@ export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) =
...
+      nonce = confirmedTransactions.transactionsList[0].nonce + 1;
+    }
+    const currentPendingNonce =
+      (stxPendingTxData?.pendingTransactions ?? []).reduce(
+        (maxNonce, transaction) => Math.max(maxNonce, transaction?.nonce ?? 0),
+        0,
+      ) + 1;
+
+    return Math.max(currentPendingNonce, nonce);
+  };
```
teebszet, line 54: this is logic that should be going into xverse-core and unit tested, because it is generic enough to be reusable and not limited to swaps.

make it a pure function (no side effects, no fetching), so it's easier to unit test.

see this PR for example of a unit tested pure function: https://github.com/secretkeylabs/xverse-core/pull/205/files

src/app/screens/swap/useAlexSponsoredTransaction.ts
```diff
@@ -24,8 +34,35 @@ export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) =
...
+
+  useEffect(() => {
+    (async () => {
+      const currentNonce = await getAccountCurrentNonce();
+      const filteredStxPendingTxData = stxPendingTxData?.pendingTransactions.filter(
+        (transaction) => currentNonce - transaction?.nonce === 1,
```
teebszet, line 60: likewise, think about where this logic should belong.
is it likely to be reused in future? then put it in core
is it specific to ALEX's sponsor service? then leave it here, but it needs to be more clear why we filter like this. you can either use a concise comment, or better naming

src/app/screens/swap/useAlexSponsoredTransaction.ts
Imamah-Zafar: thanks for suggesting the core function🙏 . I refactored the logic to filter out dead pending transactions to a simpler one. Please take a look

src/app/screens/swap/useAlexSponsoredTransaction.ts
```diff
@@ -19,11 +14,9 @@ const useAlexSponsorSwapEnabledQuery = (alexSDK: AlexSDK) =>
 export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) => {
   const alexSDK = useRef(new AlexSDK()).current;
   const [isServiceRunning, setIsServiceRunning] = useState(false);
-  const [hasPendingTransactions, setHasPendingTransactions] = useState(false);
   const { error, data: isEnabled, isLoading } = useAlexSponsorSwapEnabledQuery(alexSDK);
-  const { stxAddress } = useWalletSelector();
-  const selectedNetwork = useNetworkSelector();
-  const { data: stxPendingTxData } = useStxPendingTxData();
+  const { stxNonce } = useWalletSelector();
+  let hasPendingTransactions = false;
```
teebszet, line 19: should keep this initialisation together with the logic. (move this line down to L35)

src/app/screens/swap/useAlexSponsoredTransaction.ts
```diff
@@ -34,36 +27,16 @@ export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) =
...
-        (transaction) => currentNonce - transaction?.nonce === 1,
-      );
-
-      setHasPendingTransactions(filteredStxPendingTxData?.length > 0);
-    })();
-  }, [stxPendingTxData]);
-
+  const { data: stxPendingTxData } = useStxPendingTxData();
+  const pendingTransactionNonce =
+    stxPendingTxData?.pendingTransactions && stxPendingTxData?.pendingTransactions[0]?.nonce + 1;
```
teebszet, line 32: is the order of pendingTransactions here guaranteed? I would be a bit more explicit and take the highest nonce from pending transactions, if that should be the logic.

also the variable name pendingTransactionNonce would be more accurate if you moved the `+ 1` down to L35

src/app/screens/swap/useAlexSponsoredTransaction.ts
```diff
@@ -25,8 +28,15 @@ export const useAlexSponsoredTransaction = (userOverrideSponsorValue: boolean) =
     alexSDK.broadcastSponsoredTx(signed.serialize().toString('hex'));
 
   const { data: stxPendingTxData } = useStxPendingTxData();
-  const hasPendingTransactions = stxPendingTxData?.pendingTransactions?.length > 0;
+  const pendingTransactionNonce =
+    stxPendingTxData?.pendingTransactions && stxPendingTxData?.pendingTransactions[0]?.nonce + 1;
 
+  //ignore pending transactions if account nonce has advanced pass the nonce in pending transactions
```
teebszet, line 34: 👏

src/app/screens/swap/useAlexSponsoredTransaction.ts
Imamah-Zafar: You’re right about the pending transaction order . I have updated it so the maximum nonce value is catered for when there are multiple pending transactions in different order. 

Although I’m confused as to how adding +1 to line 35 would make a difference? The current nonce would be one value greater than the previous nonce broadcasted. So if you made a transaction with nonce 19 then the user nonce would be 20 since 20 will set as the value for the next transaction. If you calculate the nonce from the confirmed transactions (which we don’t do anymore since we simply get the value from store) but if you have to, you would get the latest confirmed transaction and add one to it. So to get the current pending tx nonce we also do the same that get the pending transaction nonce and add one to it. 
If we don’t then in case of an actual pending transaction, the stxnoxce is 20 and the pending tx nonce will be 19 and since stx nonce>pendingTransactionNonce the pending state would be ignored even though it should not have been

teebszet: thanks for the max nonce fix.

regarding the variable name: it's a case of semantics and readability. not a merge blocker, but to clarify what I meant:

> The current nonce would be one value greater than the previous nonce broadcasted.
actually there is no 'current nonce', just like there is no 'current transaction'. 
there is the previous nonce (previous transactions) and the possible 'next nonce' for future transactions. stxNonce is 'possible next nonce'

> So if you made a transaction with nonce 19 then the user nonce would be 20 since 20 will be set as the value for the next transaction.
the user nonce is not a thing. we calculate 20 but it could be rejected if another transaction gets broadcast with 20 first.

> If you calculate the nonce from the confirmed transactions (which we don’t do anymore since we simply get the value from store) but if you have to, you would get the latest confirmed transaction and add one to it.
yes, this would be determining the possible next nonce

teebszet (dismissed):

src/app/screens/swap/useAlexSponsoredTransaction.ts
Imamah-Zafar: totally agree with what you're saying. I have updated the variable name from `pendingTransactionNonce` to `upcomingPendingTransactionNonce` for readability sake

teebszet (approved):

DuskaT021:
@Imamah-Zafar can you point @blehitsahmad to the testing wallet, the one that was mentioned in the linear ticket we don't have

PR #571, opened 2023-08-16
feat: add inscription confirmation for sats connect

victorkirov:
# ⚠️ Note:
- [x] includes changes from #577 so that PR would be better merged in first for the diff to take it into account
- [ ] requires sats-connect and core PRs to be merged and deployed
- [x] ~~there is 1 todo left with the preview icon for HTML~~

# 🔘 PR Type

- [x] Enhancement

# 📜 Background
We want to be able to create inscriptions via our inscription service using Sats Connect. This adds the confirmation screens for the inscription creation.

Issue Link: ENG-2622

# 🔄 Changes
- adds the inpage, content and popup handling of the create inscription Sats Connect events
- adds a Review Inscription screen where the user can confirm their inscription details and set fee rates
- upgrades core to a version that has the functions to create inscriptions

Impact:
- This allows third party developers to create inscriptions via our inscription service by making use of the inscription functions in Sats Connect. It is best tested with the Sats Connect Example app with the inscription feature implemented.
 
# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


package.json
```diff
@@ -52,7 +52,7 @@
     "redux": "^4.0.5",
     "redux-persist": "^6.0.0",
     "redux-state-sync": "^3.1.4",
-    "sats-connect": "1.0.0",
+    "sats-connect": "^0.5.0-50cf653",
```
victorkirov, line 55: Note: pending deployment of the 2 deps

src/pages/Popup/index.tsx
```diff
@@ -26,6 +26,10 @@ const renderApp = async () => {
     buster: VERSION,
   });
   const container = document.getElementById('app');
+
+  // We need to do this after load in order for the extension popup to have the correct height on first load
+  container!.style.maxHeight = '100vh';
+
```
teebszet, line 32: @victorkirov this was part of another PR right? maybe we can add a note on the merge order to make this PR easier to review?

victorkirov: Ah, yes. Sorry, will add that 👍

teebszet: thanks! actually, I guess it doesn't help a whole lot 😂

src/common/types/inpage-types.ts
```diff
@@ -10,6 +10,7 @@ export enum DomEventName {
   signPsbtRequest = 'SatsPsbtRequest',
   signMessageRequest = 'SatsSignMessage',
   sendBtcRequest = 'SatsSendBtcRequest',
+  createInscriptionRequest = 'createInscriptionRequest',
```
abdulhaseeb4239, line 13: why did we not follow the pattern and assign 'SatsCreateInscriptionRequest'

victorkirov: Ah, good catch. I didn't notice the pattern. Thanks 😄

src/app/screens/createInscription/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+                      value={commitValue}
+                      displayType="text"
+                      thousandSeparator
+                      suffix=" sats"
+                    />
+                    <NumericFormat
+                      value={fiatValue}
+                      displayType="text"
+                      thousandSeparator
+                      suffix=" USD"
```
abdulhaseeb4239, line 368: This should not be hardcoded I think, but rather picked up from selected fiat currency. 
`const { fiatCurrency } = useWalletSelector();`

src/app/screens/createInscription/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+                        value={totalFee}
+                        displayType="text"
+                        thousandSeparator
+                        suffix=" sats"
+                      />
+                      <NumericFormat
+                        value={fiatFees}
+                        displayType="text"
+                        thousandSeparator
+                        suffix=" USD"
```
abdulhaseeb4239, line 436: same here, selected fiat currency should be rendered

abdulhaseeb4239 (changes requested):

src/app/screens/createInscription/index.tsx
victorkirov: Ah, woops! Thanks for catching this 🙏

src/app/screens/createInscription/index.tsx
```diff
@@ -0,0 +1,487 @@
...
+                </CardRow>
+              </>
+            )}
+          </CardContainer>
+          <Button onClick={onAdvancedSettingClick}>
+            <ButtonImage src={SettingIcon} />
+            <ButtonText>{t('EDIT_FEES')}</ButtonText>
+          </Button>
+          {showFeeSettings && (
+            <EditFee
```
abdulhaseeb4239, line 459: why can't we use the `TransactionSettingAlert` for the Edit fee. as it is used in other transaction screens?

abdulhaseeb4239 (dismissed):

src/app/screens/createInscription/index.tsx
victorkirov: That component handles so many different scenarios like BTC payment, ordinal payment, stacks payment. I didn't want to add another method in there and confuse the logic even further. That component needs a rewrite to separate all those into separate components, so I am using a custom local edit fees component for now.

src/app/screens/createInscription/ContentLabel/index.tsx
```diff
@@ -0,0 +1,236 @@
+import axios from 'axios';
+import { MouseEvent as ReactMouseEvent, useEffect, useMemo, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import styled from 'styled-components';
+
+import DotsVerticalIcon from '@assets/img/dots_three_vertical.svg';
+import EyeIcon from '@assets/img/eye.svg';
+import ShareIcon from '@assets/img/share.svg';
```
teebszet, line 8: FYI we are in the middle of migrating to phosphor icon set, see here for usage:
https://github.com/secretkeylabs/xverse-web-extension/blob/069859a25e37c2dbbaabc93a151be1dcc45189fc/src/app/components/seedPhraseInput/index.tsx#L69

victorkirov: 🙌  That's awesome. Will update it now.

yknl (approved):

PR #573, opened 2023-08-17
Re-enable btc recovery option

Imamah-Zafar:
# 🔘 PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
The BTC recovery option was disabled a while back and now it needs to be reenabled. 

Issue Link: #[[ENG-2662](https://linear.app/xverseapp/issue/ENG-2662/re-enable-btc-recovery-option-in-extension)]
Context Link (if applicable):

# 🔄 Changes

- Enables the Recover BTC option in recover assets screen
- Slight refactoring of the `useNonOrdinalUtxos` hook

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/1e2d9b28-5708-4e67-bffd-d7cb0d5347ba

# ✅ Review checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@ledgerhq/hw-transport-webusb": "^6.27.13",
     "@react-spring/web": "^9.6.1",
-    "@secretkeylabs/xverse-core": "1.6.1-a184dd9",
+    "@secretkeylabs/xverse-core": "1.6.1",
     "@stacks/connect": "^6.10.2",
     "@stacks/encryption": "4.3.5",
     "@stacks/stacks-blockchain-api-types": "^6.1.1",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@ledgerhq/hw-transport-webusb": "^6.27.13",
     "@react-spring/web": "^9.6.1",
-    "@secretkeylabs/xverse-core": "1.6.1",
+    "@secretkeylabs/xverse-core": "1.6.1-a184dd9",
     "@stacks/connect": "^6.10.2",
     "@stacks/encryption": "4.3.5",
     "@stacks/stacks-blockchain-api-types": "^6.1.1",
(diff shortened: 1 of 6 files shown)


src/app/screens/restoreFunds/restoreBtc/index.tsx
```diff
@@ -130,7 +130,7 @@ function RestoreBtc() {
   const onClickTransfer = () => {
     mutateSignNonOrdinalBtcTransaction({
       recipientAddress: btcAddress,
-      nonOrdinalUtxos: unspentUtxos,
+      nonOrdinalUtxos: unspentUtxos ?? [],
```
victorkirov, line 133: I don't think we really want to fire this event if unspentUtxos is undefined. Maybe something like this rather?

```suggestion
const onClickTransfer = () => {
  if (!unspentUtxos) return;
    mutateSignNonOrdinalBtcTransaction({
      recipientAddress: btcAddress,
      nonOrdinalUtxos: unspentUtxos,
```

package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@ledgerhq/hw-transport-webusb": "^6.27.13",
     "@react-spring/web": "^9.6.1",
-    "@secretkeylabs/xverse-core": "1.6.1",
+    "@secretkeylabs/xverse-core": "1.6.1-a184dd9",
```
victorkirov, line 9: Out of interest, which core PR is this?

src/app/screens/restoreFunds/restoreBtc/index.tsx
victorkirov: We might also want to disable the button and show a loader while unspentUtxos is undefine

package.json
Imamah-Zafar: This was not relevant to the PR. The core version has been set back to 1.6.1. I was under the impression that the btc recovery feature was added in the beta version of core since it was added in the app side PR but talked with haseeb and it was added since it was for an app specific fix. Reverted the change in the PR

victorkirov: Ah, awesome :)

victorkirov (approved):
LGTM

yknl:
Transferring of BTC from the ordinals address has been enabled in release v0.20.0.

PR #577, opened 2023-08-24
fix: full screen and popup styling

victorkirov:
# 🔘 PR Type

- [x] Other... Please describe: CSS

# 📜 Background
If you open the popup via sats connect and the browser was in full screen mode, it would open the popup in full screen and place it in the top left corner. If you were not in full screen, it would open the popup in it's own window that was the max height for an extension (600px) but that height was the outer height (including the title bar) resulting in an inner height of 572px which would make the contents a bit bigger than the window resulting in an annoying scrolling behaviour.

This makes the full screen window centre the contents on screen, and fixes the scroll issue.

Issue Link: ENG-2707

# 🔄 Changes
Impact:
- changes is styling on the main app container, so it would effectively change the outer styling for all popups and pages.

# 🖼 Screenshot / 📹 Video
Before:

https://github.com/secretkeylabs/xverse-web-extension/assets/3322667/337da45b-036d-4181-a35c-8f881ea0c764

After:

https://github.com/secretkeylabs/xverse-web-extension/assets/3322667/34a261c5-0c38-43fa-bd49-57ace0696883

# ✅ Review checklist


teebszet (dismissed):

teebszet:
https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/49e60a74-3c23-4f27-ae87-70232fc1e1ab

src/app/components/screenContainer/index.tsx
```diff
@@ -6,7 +6,7 @@ import styled from 'styled-components';
 const RouteContainer = styled.div((props) => ({
   display: 'flex',
   flexDirection: 'column',
-  height: 600,
+  height: '100%',
```
teebszet, line 9: https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/109fa345-8735-4b1a-9dd9-1cbe663d5469

RouteContainer here is used also by full screen /options routes

victorkirov:
> Screen.Recording.2023-08-24.at.4.52.22.PM.mov

Ah, I didn't test the onboarding 😅 Thanks, will sort it out now.

src/app/components/screenContainer/index.tsx
victorkirov: Should be sorted 👍 Thanks for catching

src/pages/Options/index.css
```diff
@@ -28,3 +28,7 @@ header h2 {
   justify-content: center;
   align-items: center;
 }
+
+#app {
+  min-height: 600px;
```
victorkirov, line 33: @teebszet 

This is the fix for the onboarding page. With min-height, there is the possibility of it extending past 600 if the contents are bigger, instead of changing over to scrollbar. I couldn't find a screen that's affected by it though.

An alternative would be to add a context provider in the layouts which sets the context to popup or full screen in the provider, and we can add a consumer on the screenContainer component which either sets the height to 600px or 100%, depending on the context?

teebszet: I think the simpler solution is better here. thanks for fixing!

teebszet (approved):

DuskaT021:
this is tested and it can be merged @teebszet

PR #581, opened 2023-08-28
feat: add Xverse namespaced providers

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
Currently we only inject the default stacks and sats providers into the inpage window object. This clashes with other wallets which also inject it in the same place and integrators cannot force select Xverse when calling stacks/sats connect.

Issue Link: ENG-2732

# 🔄 Changes
This adds default Xverse providers in an Xverse namespace object on the window object.

# ✅ Review checklist


teebszet (approved):

DuskaT021:
This can be merged 👍

PR #582, opened 2023-08-28
Add scrollbar on all scrollable screens

Imamah-Zafar:
# 🔘 PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
The scrollbar design needs to be updated. Currently, the scrollbar is hidden and is not visible. We want to show the scrollbar on hover so user knows the screen is scrollable

Issue Link: #[[ENG-2337]](https://linear.app/xverseapp/issue/ENG-2337/scrollbar-on-all-scrollable-screens)
Context Link (if applicable):

# 🔄 Changes
Keeping in mind the impact of the changes, A component `ScrollBar` is introduced in the PR and is added to the home screen. The scrollbar will be included in the rest of the screens gradually in the UI revamp. Rest of the screens have the old transparent scrollbar currently

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/0e8ae223-46b1-4e67-94c2-0339b64240c9

# ✅ Review checklist


src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+    maxHeight: 10,
+    borderRadius: 24,
+    background: 'var(--white-800, rgba(255, 255, 255, 0.20))',
+  },
+});
+
+interface Props {
+  children: ReactNode;
+}
+function ScrollBar({ children }: Props): JSX.Element {
```
teebszet, line 26:
```suggestion
function ScrollBar({ children }: React.PropsWithChildren<{}>): JSX.Element {
```

src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  '&::-webkit-scrollbar': {
+    width: 8,
+    height: 10,
+    background: 'transparent',
+  },
+  '&::-webkit-scrollbar-thumb': {
+    width: 8,
+    maxHeight: 10,
+    borderRadius: 24,
+    background: 'var(--white-800, rgba(255, 255, 255, 0.20))',
```
teebszet, line 19: is --white-800 defined anywhere?

why don't we use the props.theme.colors.white_800 and props.theme.colors.white_600?

src/app/screens/swap/index.tsx
```diff
@@ -44,10 +41,10 @@ const DownArrowButton = styled.button((props) => ({
 }));
 
 const SendButtonContainer = styled.div((props) => ({
-  paddingBottom: props.theme.spacing(12),
+  paddingBottom: props.theme.spacing(42),
   paddingTop: props.theme.spacing(4),
-  marginLeft: '5%',
-  marginRight: '5%',
+  marginLeft: props.theme.spacing(8),
+  marginRight: props.theme.spacing(4),
```
teebszet, line 47: what's the reasoning for not keeping the scrollbar as display:none? it's going to be really brittle if all code which uses ScrollBar has to explicitly allow the right margin.

src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+const ScrollBarContainer = styled.div({
+  overflowY: 'hidden',
+  scrollbarGutter: 'stable',
+  ':hover': {
+    overflowY: 'auto',
+  },
+  '&::-webkit-scrollbar': {
+    width: 8,
+    height: 10,
+    background: 'transparent',
```
teebszet, line 13: why are we making the scrollbar transparent and taking up width, rather than display:none ?

src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { ReactNode } from 'react';
+import styled from 'styled-components';
+
+const ScrollBarContainer = styled.div({
+  overflowY: 'hidden',
+  scrollbarGutter: 'stable',
+  ':hover': {
+    overflowY: 'auto',
+  },
```
teebszet, line 9: what's the reasoning for changing the overflow on hover?

src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+function ScrollBar({ children }: Props): JSX.Element {
+  return (
+  <ScrollBarContainer>
+    {children}
+  </ScrollBarContainer>
+  );
+}
+
+export default ScrollBar;
```
teebszet, line 34: I'm not sure using a react container is the best way to do this.

if all we need to do is centralise the hidden scrollbars CSS, then let's follow the same pattern as in theme/index.ts with typographies.

then usage will be same as for apply typography styles to any styled component, and you don't need to refactor all the JSX which wants hidden scrollbars

src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  },
+  '&::-webkit-scrollbar': {
+    width: 8,
+    height: 10,
+    background: 'transparent',
+  },
+  '&::-webkit-scrollbar-thumb': {
+    width: 8,
+    maxHeight: 10,
+    borderRadius: 24,
```
teebszet, line 18: is there a way to do this without the scrollbar taking up width? then you wouldn't have to update all the marginRight on each consuming component

teebszet:
am I missing something in the video? 😅  there are no scrollbars on hover

src/app/components/tabBar/index.tsx
```diff
@@ -16,12 +16,21 @@ const RowContainer = styled.div((props) => ({
...
   borderTop: `1px solid ${props.theme.colors.background.elevation3}`,
 }));
 
+const BottomBarContainer = styled.div({
+  position: 'absolute',
+  bottom: 0,
+  left: 0,
+  right: 0,
+});
+
```
teebszet, line 33: this breaks a lot of layouts because now the other components will fill up the space vacated by changing this to absolute positioning. what was the reasoning here? I would strongly suggest not doing this. look at position: fixed and position: sticky as alternatives

for example, here the scrollbar appears behind the bottom tab, and the bottom tab is covering the 'Manage Token List' button: 
image

src/app/components/scrollBar/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { ReactNode } from 'react';
+import styled from 'styled-components';
+
+const ScrollBarContainer = styled.div({
+  overflowY: 'hidden',
+  scrollbarGutter: 'stable',
```
teebszet, line 6: consider using stable both-edges?

Imamah-Zafar:
> am I missing something in the video? 😅 there are no scrollbars on hover

aah sorry my bad. i added the wrong video by mistake . I'll update it after addressing the comments

src/app/components/scrollBar/index.tsx
Imamah-Zafar: OK I'm currently working on only adding the scrollbar CSS rather than a separate component. However, the scrollable screens do need one refactoring even with the CSS which would be around the margin. since we need to use `scrollbar-gutter: stable both edges` property which allocates symmetric spacing to both sides of the screen so that the layout does not shift when the scroll bar is visible . The width of the scrollbar gutter is the same as the width of the scrollbar so in our case 8px. This would mean there would be a 8px margin on both sides on all screens

teebszet: yep, I think that's still preferable to refactoring with asymmetric left/right margins

src/app/components/tabBar/index.tsx
teebszet: successfully pushed back on design for now

src/app/components/extendedScreenContainer/index.tsx
```diff
@@ -1,17 +1,18 @@
...
+interface Props {
+  isGalleryOpen: boolean;
+}
+const ExtendedScreenRouteContainer = styled.div<Props>((props) => ({
   display: 'flex',
   flexDirection: 'column',
-  height: '100vh',
-  width: '100vw',
+  height: props.isGalleryOpen ? '100vh' : '100%',
+  width: props.isGalleryOpen ? '100vw' : '100%',
```
teebszet, line 14: will want to test this doesn't break all screens which use ExtendedScreenRouteContainer

Imamah-Zafar:
@teebszet After merging develop there were a bunch (20+) eslint errors that I had to resolve in the PR in order to push my changes. 
updated the PR according to our discussion to limit the scope of the PR. please review

src/app/components/transactionsRequests/ContractCallRequest.tsx
```diff
@@ -147,7 +147,7 @@ export default function ContractCallRequest(props: ContractCallRequestProps) {
   const getFunctionArgs = (): Array<ArgToView> => {
     const args: Array<ArgToView> = [];
     const { funcArgs } = extractFromPayload(request);
-    funcMetaData?.args?.map((arg: Args, index: number) => {
+    funcMetaData?.args?.forEach((arg: Args, index: number) => {
```
teebszet, line 150: hey, thanks for fixing these eslint errors, but in the spirit of reducing QA test scope for this PR, let's not make functional changes like this without having time to really test the implications.

src/app/components/transactionsRequests/ContractCallRequest.tsx
```diff
@@ -269,23 +269,28 @@ export default function ContractCallRequest(props: ContractCallRequestProps) {
 
   const renderPostConditionsCard = () => {
     const { postConds } = extractFromPayload(request);
+    let coinInfo;
     return postConds?.map((postCondition, i) => {
       switch (postCondition.conditionType) {
         case PostConditionType.STX:
-          return <StxPostConditionCard key={i} postCondition={postCondition} />;
+          return <StxPostConditionCard key={postCondition.type} postCondition={postCondition} />;
```
teebszet, line 276: i'm not 100% sure on what postCondition.type is, but the react key should be an identifier, unique within the list. if it's possible or likely for the array to have more than one item with the same postCondition.type, then it's not a good choice for the key

src/app/screens/login/index.tsx
```diff
@@ -22,12 +22,20 @@ const Logo = styled.img({
...
 const ScreenContainer = styled(animated.div)({
   display: 'flex',
   flexDirection: 'column',
   flex: 1,
   paddingLeft: 18,
   paddingRight: 18,
+  overflowY: 'auto',
+  '&::-webkit-scrollbar': {
+    display: 'none',
+  },
```
teebszet, line 38: what was the reason for this?

src/app/screens/restoreFunds/restoreBtc/index.tsx
```diff
@@ -104,25 +104,25 @@ function RestoreBtc() {
       recipientAddress: string;
       nonOrdinalUtxos: Array<UTXO>;
       accountIndex: number;
-      seedPhrase: string;
-      network: NetworkType;
+      userSeedPhrase: string;
```
teebszet, line 107: probably don't need to update this, as we have an open PR for seed vault which will be touching this

src/theme/index.ts
```diff
@@ -192,6 +192,24 @@ const Theme = {
...
 
+  scrollbar: {
+    scrollbarGutter: 'stable both-edges',
+    'overflow-y': 'hidden',
+    ':hover': {
+      'overflow-y': 'auto',
+    },
+    '::-webkit-scrollbar': {
+      width: 8,
+      background: 'var(--white-900, rgba(255, 255, 255, 0.10))',
```
teebszet, line 203: I don't think we have CSS vars defined anywhere with --white-900. it should be removed
```suggestion
      background: 'var(rgba(255, 255, 255, 0.1))',
```

src/app/screens/login/index.tsx
Imamah-Zafar: without this the default chrome scrollbar was being shown. Currently, files other than home already have this in their css, which is why no scrollbar gets shown on those screens. This was not required before since this CSS was also added in the `pages/Popup/index.css` which had to be removed in order to show the visible scrollbar 

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/4c4a9682-3f92-4faa-8ec4-986e6f16c907

teebszet: ah ok, got it

teebszet (approved):

blehitsahmad:
@Imamah-Zafar This is good to merge. As expected, the scroll bar is visible only on the main dashboard. I have verified the scroll functionality as well. 

The following screens were tested:

- Dashboard
- Accounts
- Transaction History
- Settings
- Send asset
- Manage token list
- Send button
- Collectables

**Browsers used**: Brave, Chrom

Imamah-Zafar:
Noticed the hidden scrollbar css was missing in two screens which i added in the latest commit 

@teebszet need a re review and then we can merge

teebszet (approved):

PR #584, opened 2023-08-28
Fix Tx history screen reload & send form navigation issues

Imamah-Zafar:
# 🔘 PR Type

- [x] Bugfix
- [ ] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
Transaction history page for all transaction reloads every couple of seconds showing a spinner every time which is not the best UX. 
There were also navigation issues while fund transfer. 
On pressing the back button on the STX send form goes back to the confirmation screen, which is incorrect behavior.
The input fields also get cleared when user navigates back from confirmation screen to send screen

Issue Link: #[[ENG-2678](https://linear.app/xverseapp/issue/ENG-2678/navigation-issue-when-sending-stx)]
              [ENG-2588](https://linear.app/xverseapp/issue/ENG-2588/transaction-history-page-for-all-transaction-reloads-every-couple-of)

# 🔄 Changes

- The transaction history UX is improved so that the spinner is not shown on refetch also the screen does not scroll up every time the reloading happens. 
- The navigation issue in the STX flow is fixed where  the user clicks the back button on the send screen you get redirected to the home page
- The clearance of input fields when the user navigates back was introduced when the check was added to clear funds transfer progress if the account is switched. The useEffect was being called on the first render hence the input fields would get cleared if the user navigates backward from the confirmation screen to the send screen. The PR fixes the bug by creating a customer hook `useClearFormOnAccountSwitch`. The hook sees if its the first render then doesn't clear the input fields. 

# 🖼 Screenshot / 📹 Video

### Before

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/f03a9410-98d2-4593-962b-4c4d732badeb

### After

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/27ab8aea-6801-48d2-89b7-1f0c87d7a3f0

## Navigation issues

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/8e3ec365-0667-4d98-beea-cba0abb83ef8

## Change account flow 

https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/c38ab77d-0689-41ed-9d60-e090511c23cc

# 🧪 Testing Instructions 
- Verify the transaction history screen reload issue is fixed where the spinner and jump to top does not happen after every couple of seconds 
- Try the STX/BTC send flow and see if you go back from the confirm screen to send screen the input fields are populated.
- On pressing the back button on the STX send form goes back to the main dashboard
- Also confirm the flow where if you are in the middle of filling the form and the account is changed then the expected behavior should be that the input fields should be cleared. 

# ✅ Review checklist


src/app/components/sendForm/useClearFormOnAccountSwitch.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  const isFirstRender = useRef(false);
+  useEffect(() => {
+    if (isFirstRender.current) {
+      setIsAccountSwitched(true);
+    } else {
+      //if first render then dont clear form input fields
+      isFirstRender.current = true;
+    }
+  }, [selectedAccount]);
```
teebszet, line 16: @Imamah-Zafar nice job with the hook, I like the abstraction.

just wondering, did you test this with multiple account switches? after isAccountSwitched is set to true once, it looks like it won't trigger again

src/app/screens/sendStx/index.tsx
```diff
@@ -73,7 +73,7 @@ function SendStxScreen() {
   }, [data]);
 
   const handleBackButtonClick = () => {
-    navigate(-1);
+    navigate('/');
```
teebszet, line 76: back buttons should all navigate(-1), because we don't assume the sendStx screen has only one entrypoint. recently there was another bug to fix the accountHeader back buttons.

if the bug is the looping back from: https://linear.app/xverseapp/issue/ENG-2678/navigation-issue-when-sending-stx, then probably the confirmStx screen needs to navigate(-1) instead of navigate(/send-stx). 

you can think of the navigation state as a stack, where the back button should be popping off the top instead of putting another on

Imamah-Zafar: the reason why navigate(/send-stx) was set since we want to pass the state from confirm screen to send screen. If you set navigate(-1) then from my understanding, the previous state cannot be fetched. In the documentation it's written that

src/app/components/sendForm/useClearFormOnAccountSwitch.ts
Imamah-Zafar: yes you're right. It wasn't being triggered for multiple account switches, I updated the dependency array for the useEffect so it would get called when the account changes . It's working now for multiple switches 



https://github.com/secretkeylabs/xverse-web-extension/assets/88320460/02ba60fb-4fec-4d83-9d3d-faeee03823a5

src/app/screens/sendStx/index.tsx
teebszet: ok, thanks for looking into that. 

can you put a very short comment in the code to give some context here, explaining why we take this shortcut, even if it's less robust

teebszet (approved):

teebszet (dismissed):

blehitsahmad:
@Imamah-Zafar if the user adds a memo when performing an STX transaction and proceeds, then hits back to edit the form again, the user gets the "memo exceeds the allowed length" error. This error is also present on prod; however, this could be fixed under the scope of this PR. 

https://github.com/secretkeylabs/xverse-web-extension/assets/108722927/fb511339-051a-45d8-a094-9952eca8ddbf

blehitsahmad:
@Imamah-Zafar This is good to merge - the memo issue has been fixed as well. 

Following test cases (high level) were executed:
# 

- [ ]  Check STX screen for reload issue (scroll to verify there’s no pagination issues)
- [ ]  Verify that all fields remain populated when the user hits back
- [ ]  Verify that only the fields that were populated remain filled when the user hits back
- [ ]  Verify that upon account switching the fields are reset
- [ ]  Verify that clicking cancel does not reset the fields
- [ ]  Verify that the memo issue is fixed (proceed with no memo - back - proceed as it is - back - change memo - proceed - back - proceed again - back - change memo - proceed)
- [ ]  Proceed - cancel - back (verify that the user is not stuck in a loop between the two screens)

teebszet:
@Imamah-Zafar needs conflicts resolved

Imamah-Zafar:
> @Imamah-Zafar needs conflicts resolved

done 👍

teebszet (approved):

PR #585, opened 2023-08-29
feat: 1 step brc 20 transfers

teebszet:
# 🔘 PR Type
- [x] Enhancement

# 📜 Background
Issue Link: https://linear.app/xverseapp/issue/ENG-2593/implement-new-send-brc-20-flow 

and revamp tasks: 
https://linear.app/xverseapp/issue/ENG-2719/implement-colour-changes
https://linear.app/xverseapp/issue/ENG-2495/implement-typography-changes

**This PR should maintain the existing support for ledger 2-step brc20 flow**

NOT included in this PR :
- (still finalising design) https://linear.app/xverseapp/issue/ENG-2785/notice-for-an-non-taproot-address-when-sending-brc-20-tokens
- ledger 1 step brc20 flow https://linear.app/xverseapp/issue/ENG-2786/ledger-1-step-brc-20-transfers-web-extension

# 🔄 Changes
functional changes:
- updated screen: send brc20 now includes recipient input
- new screen: confirm brc20 now shows details and edit fees according to figma
- new screen: execute brc20 is the loading page with animation showing progress
- updated component: account header and account select no longer show the 'copy address' buttons

codebase changes:
- new component: LoadingTransactionStatus is a generic component (main credit: @fedeerbes)
- new component: BetterBarLoader is react-content-loader which takes width and height as props
- new components: common.styled.ts is a place for reusable styled components (rules defined by zeroheight)
- updated theme colors: updated elevation colours to revamped and changed usage to be a flat structure under theme.colors.
- updated typography usage: updated typography font size, weight, family to revamped and changed usage to be a flat structure under theme.typography
- refactored useResetUserFlow to simplify usage

Impact:
- 1 step brc20 flow from the coin dashboard entry point
- if using ledger account, brc20 flow should still be the old 2-step flow
- account headers
- some minor styling and react warning fixes on: 
  - bottom tab bar
  - account headers
  - account select rows
  - token image
  - recipient component

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/3b13429e-775c-40cd-9bec-f097228804af

account select rows check here:
image

ledger account still supports 2-step brc20 flow:

https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/e51b74c3-7166-4f53-8217-ce2be69eee4f

# ✅ Review checklist


src/app/components/confirmBtcTransactionComponent/index.tsx
```diff
@@ -113,7 +113,7 @@ const ReviewTransactionText = styled.h1<ReviewTransactionTitleProps>((props) =>
 
 interface Props {
   currentFee: BigNumber;
-  feePerVByte: BigNumber;
+  feePerVByte: BigNumber; // is this the same as currentFeeRate?
```
victorkirov, line 116: This is a strange setup. It looks like feePerVByte is the initial value received from the query params and is used to initialise the currentFeeRate in this component and in the fee editor.

Maybe we should refactor this? Out of scope for this PR bnut maybe change the comment into a `TODO`?

src/app/components/loadingTransactionStatus/circularSvgAnimation.tsx
```diff
@@ -0,0 +1,110 @@
...
+        />
+        <g className={crossClass}>
+          <path d="M55 33L33 55" />
+          <path d="M55 55L33 33" />
+        </g>
+      </StyledSvg>
+    </animated.div>
+  );
+}
+export default CircularSvgAnimation;
```
victorkirov, line 110: 🪄 Very cool!

src/app/components/sendForm/index.tsx
```diff
@@ -160,6 +160,7 @@ const SendButtonContainer = styled.div<ButtonProps>((props) => ({
   opacity: props.enabled ? 1 : 0.6,
 }));
 
+/* --------- move to separate file ------------- */
```
victorkirov, line 163: Do you want to do this as part of this PR? If not, please suffix with TODO

src/app/components/tabBar/index.tsx
```diff
@@ -60,22 +60,26 @@ function BottomTabBar({ tab }: Props) {
...
-      <Button onClick={handleNftButtonClick}>
-        <img src={tab === 'nft' ? NftTab : UnselectedNftTab} alt="nft" />
-      </Button>
-      <Button onClick={handleStackingButtonClick}>
-        <img src={tab === 'stacking' ? StackingTab : UnselectedStackingTab} alt="stacking" />
-      </Button>
-      <Button onClick={handleSettingButtonClick}>
-        <img src={tab === 'settings' ? SettingsTab : UnselectedSettingsTab} alt="settings" />
-      </Button>
+      {showBottomBar && (
```
victorkirov, line 65: I think this changes the functionality a bit. before, if showBottomBar was falsey, it would return null. Now, it will always render the RowContainer which has a minHeight of 64px, so it would render an empty div. Is that the expected behaviour?

src/theme/index.ts
```diff
@@ -16,17 +52,18 @@ const Theme = {
...
+      elevation_1: '#1E2024', // deprecated
+      elevation2: '#24282F',
+      elevation3: '#2A2F39',
+      elevation5: '#303643',
+      elevation6: '#4C525F',
+      elevation6_600: 'rgba(76, 82, 95, 0.4)',
+      elevation6_800: 'rgba(76, 82, 95, 0.2)',
+      elevation8: '#7A7688',
+      elevation9: 'rgba(76, 81, 135, 0.2)', // deprecated
+      elevation10: 'rgba(76, 81, 135, 0.35)', // deprecated
```
victorkirov, line 66: Were these supposed to change in this PR? Looks like you added the new ones above; is the plan not to migrate from these to those but keep these with the old values for now? Just checking because it seems like a potentially riskyt change 😅

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  overflow-y: auto;
+  &::-webkit-scrollbar {
+    display: none;
+  }
+  height: 600px;
+  width: 360px;
+  margin: auto;
```
victorkirov, line 32: I don't think you need any of these.

```suggestion
```

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          recipients={[
+            { address: recipientAddress, amountSats: new BigNumber(estimateFeesParams.amount) },
+          ]}
+          token={token}
+          totalFee={totalFee}
+          transactionFee={txFee}
+          isFeeLoading={isFeeLoading}
+          error={error}
+        />
+        {!isInOptions && <BottomBar tab="dashboard" />}
```
victorkirov, line 133: This probably shouldn't be here. Also, `isInOptions` is a function, so you'd need to call it to get the boolean value.

src/app/hooks/useResetUserFlow.ts
```diff
@@ -14,6 +14,7 @@ const userFlowConfig: Record<string, { resetTo: string }> = {
   '/confirm-btc-tx': { resetTo: '/send-btc' },
   '/send-brc20': { resetTo: '/account-list' },
   '/confirm-inscription-request': { resetTo: '/account-list' },
+  '/confirm-brc20-tx': { resetTo: '/account-list' },
```
victorkirov, line 17: `/account-list` is a strange place to reset to. Should it not rather be `/` or `/send-brc20`?

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const { selectedAccount } = useWalletSelector();
+  const navigate = useNavigate();
+  const {
+    recipientAddress,
+    estimateFeesParams,
+    estimatedFees: initEstimatedFees,
+    token,
+  }: ConfirmBrc20TransferState = useLocation().state;
+  const { subscribeToResetUserFlow } = useResetUserFlow();
+  useEffect(() => subscribeToResetUserFlow('/confirm-brc20-tx'), [subscribeToResetUserFlow]);
```
victorkirov, line 46: Note that `useResetUserFlow` doesn't actually memoise `subscribeToResetUserFlow`, so each render results in the useEffect firing and the subscription being closed and re-opened. 

Also, in testing, I noticed that when this happened, the channel is closed on the useEffect rerendering, and when we run the subscribe again, we are subscribing to a closed channel because nothing opens it again. That results in the reset message never actually firing.

src/app/components/tabBar/index.tsx
teebszet: Yeah sorry didn't document this change, but it's the expected new behaviour in revamped design

src/app/screens/confirmBrc20Transaction/index.tsx
teebszet: Nice catch thanks!

src/theme/index.ts
teebszet: https://github.com/secretkeylabs/xverse-web-extension/pull/588

src/app/screens/confirmBrc20Transaction/index.tsx
victorkirov: It might actually be better to refactor that component to have a single arg of `path` and internally use a single useEffect where it creates a channel, subscribes as per subscribeToResetUserFlow with the input path, and returns the close channel function, that way it's all contained in a single useEffect with no race conditions.

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        } else {
+          setError(BRC20ErrorCode.SERVER_ERROR);
+        }
+      }
+      setIsFeeLoading(false);
+    };
+    if (debouncedUserInputFeeRate) {
+      runEstimate();
+    }
+  }, [debouncedUserInputFeeRate, estimateFeesParams]);
```
victorkirov, line 82: 🤔 why not use the fee estimate hook from Core?

src/app/screens/confirmBrc20Transaction/confirmBrc20TransactionComponent.tsx
```diff
@@ -0,0 +1,236 @@
+import ActionButton from '@components/button';
+import BigNumber from 'bignumber.js';
+import Brc20FeesComponent from '@screens/confirmBrc20Transaction/brc20FeesComponent';
```
victorkirov, line 3: nit: For consistency, would prefer:
```suggestion
import Brc20FeesComponent from './brc20FeesComponent';
```

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ConfirmBrc20TransferState,
+  ExecuteBrc20TransferState,
+} from '@utils/brc20';
+import { isInOptions, isLedgerAccount } from '@utils/helper';
+import { useEffect, useState } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useResetUserFlow } from '@hooks/useResetUserFlow';
+import { useTranslation } from 'react-i18next';
+import ConfirmBrc20TransactionComponent from './confirmBrc20TransactionComponent';
+import { OnChangeFeeRate } from './editFees';
```
victorkirov, line 20:
```suggestion
import type { OnChangeFeeRate } from './editFees';
```

src/app/screens/confirmBrc20Transaction/confirmBrc20TransactionComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+import SettingIcon from '@assets/img/dashboard/faders_horizontal.svg';
+import TransactionDetailComponent from '@components/transactionDetailComponent';
+import styled from 'styled-components';
+import useWalletSelector from '@hooks/useWalletSelector';
+import { NumericFormat } from 'react-number-format';
+import { Recipient } from '@secretkeylabs/xverse-core/transactions/btc';
+import { FungibleToken, getBtcFiatEquivalent } from '@secretkeylabs/xverse-core';
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { EditFees, OnChangeFeeRate } from './editFees';
```
victorkirov, line 14:
```suggestion
import type { OnChangeFeeRate } from './editFees';
import { EditFees } from './editFees';
```

src/app/hooks/useResetUserFlow.ts
teebszet: there was a fix here: https://github.com/secretkeylabs/xverse-web-extension/pull/538

so I guess it could be '/', but not '/send-brc20' because that route needs some query params to not crash

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <ConfirmBrc20TransactionComponent
+          btcFee={btcFee}
+          currentFeeRate={Number(userInputFeeRate) || estimateFeesParams.feeRate}
+          inscriptionFee={inscriptionFee}
+          onClickApplyFee={handleClickApplyFee}
+          onChangeFee={handleClickApplyFee}
+          onClickCancel={handleClickCancel}
+          onClickConfirm={handleClickConfirm}
+          recipients={[
+            { address: recipientAddress, amountSats: new BigNumber(estimateFeesParams.amount) },
```
victorkirov, line 125: I think it would be better to create a new Recipient type for BRC-20 and not reuse the btc one. `amountSats` is confusing for a BRC-20 amount.

src/app/screens/confirmBrc20Transaction/index.tsx
teebszet: I wanted to use the hook from core (including the error handling), but because of the design, I need to run the estimate fee:

-  once in the send-brc20 screen to validate whether we can move to the confirm-brc20 screen
- and not on the confirm-brc20 screen on first render (because we have no user input yet), but again on change of the user input field

I couldn't control when the core hook runs because of the 'no conditional hooks' rule, hence line 79-81

victorkirov: Ah, ok, cool. Maybe we can refactor that hook somehow 🤔

src/app/screens/confirmBrc20Transaction/brc20FeesComponent.tsx
```diff
@@ -0,0 +1,96 @@
...
+`;
+
+const FiatAmountText = styled.p`
+  ${(props) => props.theme.body_medium_s}
+  color: ${(props) => props.theme.colors.white[400]};
+  display: flex;
+  justify-content: flex-end;
+`;
+
+function Brc20FeesComponent({
```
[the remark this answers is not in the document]
teebszet: 🤔 yes I guess that makes sense if I click X or cancel. will change it

src/app/screens/confirmBrc20Transaction/confirmBrc20TransactionComponent.tsx
teebszet: is there a functional difference to splitting the type out to import type like this, when you're already importing a function as well?

src/app/screens/confirmBrc20Transaction/index.tsx
teebszet: right! it makes more sense without the subscribeToResetUserFlow as an effect dep, so removed that first. I also added a TODO to refactor the useResetUserFlow, but I remember I wanted to keep it flexible in case the use case required more control

src/app/screens/confirmBrc20Transaction/editFees.tsx
```diff
@@ -0,0 +1,300 @@
...
+              value={feeRateInput?.toString()}
+              onChange={onChangeEditFeesInput}
+            />
+            <FeeText>sats /vB</FeeText>
+            <TickerContainer>
+              {isFeeLoading ? (
+                <>
+                  <BetterBarLoader width={75} height={16} />
+                  <BetterBarLoader width={75} height={16} />
+                </>
```
victorkirov, line 258: Why are there 2 of these? 😋

src/app/utils/brc20.ts
```diff
@@ -0,0 +1,57 @@
...
+  commitChainFee,
+  revealChainFee,
+  transferChainFee,
+  revealServiceFee,
+  transferUtxoValue,
+}: TransferEstimateResult['valueBreakdown']) => {
+  const txFee = new BigNumber(commitChainFee).plus(revealChainFee).plus(transferChainFee);
+  const inscriptionFee = new BigNumber(revealServiceFee);
+  const totalFee = new BigNumber(txFee).plus(inscriptionFee);
+  const btcFee = new BigNumber(transferUtxoValue);
```
victorkirov, line 54: This isn't really the btc fee. It's the sats value of the final UTXO that will hold the inscription.

src/app/screens/executeBrc20Transaction/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+  const [loadingPercentage, setLoadingPercentage] = useState(0);
+
+  useEffect(() => {
+    if (complete || errorCode) {
+      return;
+    }
+    executeTransfer();
+  }, [executeTransfer, complete, errorCode]);
```
victorkirov, line 37: I think we only want to execute executeTransfer once on load. Executing it more than once would result in an error somewhere along the way. This might be enough?

```suggestion
  useEffect(() => {
    executeTransfer();
  }, []);
```

src/app/screens/sendBrc20/index.tsx
```diff
@@ -1,91 +1,66 @@
...
-  const location = useLocation();
+  const [recipientError, setRecipientError] = useState('');
+  const [recipientAddress, setRecipientAddress] = useState('');
+  const [processing, setProcessing] = useState(false);
 
-  const coinName = location.search ? location.search.split('coinName=')[1] : undefined;
-  const fungibleToken =
-    location.state?.fungibleToken || brcCoinsList?.find((coin) => coin.name === coinName);
+  const { subscribeToResetUserFlow } = useResetUserFlow();
+  useEffect(() => subscribeToResetUserFlow('/send-brc20'), [subscribeToResetUserFlow]);
```
victorkirov, line 39: Just a note that there will be the same issue here.

src/app/screens/sendBrc20/index.tsx
```diff
@@ -94,123 +69,96 @@ function SendBrc20Screen() {
     if (!resultRegex.test(newValue)) {
       setAmountToSend('');
     } else {
+      validateAmount(newValue);
       setAmountToSend(newValue);
     }
```
victorkirov, line 74: This has a slightly annoying functionality in that if I want to type in something like `20350` and I accidentally type in `2035-`, then it resets the text to `''`.

I think either try this:
```suggestion
    if (resultRegex.test(newValue)) {
      validateAmount(newValue);
      setAmountToSend(newValue);
    }
```

or something like:
```suggestion
    const newValue = +e.currentTarget.value;
    if (!Number.isNan(newValue)) {
      validateAmount(newValue);
      setAmountToSend(newValue);
    }
```

src/app/screens/sendBrc20/index.tsx
```diff
@@ -94,123 +69,96 @@ function SendBrc20Screen() {
...
-  };
+      setProcessing(true);
+
+      if (
+        !validateAmount(amountToSend) ||
+        !validateRecipientAddress(recipientAddress) ||
+        !feeRate
+      ) {
+        return setProcessing(false);
+      }
```
victorkirov, line 110: nit:
```suggestion
      if (
        !validateAmount(amountToSend) ||
        !validateRecipientAddress(recipientAddress) ||
        !feeRate
      ) {
        return;
      }
      
      setProcessing(true);
```

src/app/screens/sendBrc20/index.tsx
```diff
@@ -1,91 +1,66 @@
...
-    seedPhrase,
-    network,
-    btcFiatRate,
-    brcCoinsList,
-  } = useWalletSelector();
+  const location = useLocation();
+  const { btcAddress, ordinalsAddress, network, brcCoinsList } = useWalletSelector();
+  const { data: feeRate } = useBtcFeeRate();
   const [amountError, setAmountError] = useState('');
   const [amountToSend, setAmountToSend] = useState('');
```
victorkirov, line 33: This would probably be easier to keep as a number instead of casting it everywhere 😋 

```suggestion
  const [amountToSend, setAmountToSend] = useState(0);
```

victorkirov (changes requested):
Initial review done 😄 Just a few small things to look into

src/app/screens/sendBrc20/index.tsx
teebszet: thanks for testing this! both seem good, I might go for door number 2 and use !Number.isFinite()

victorkirov: I usually use door #2 😋 Personally, I prefer having the useState hold a number instead of a string and I use the code in the snippet to cast it to a number and only set the state if it's valid. I guess you could use the isFinite first and only cast it to number if isFinite returns true, but I think under the hood that would be marginally more expensive as it would effectively cast it twice.

src/app/screens/confirmBrc20Transaction/index.tsx
victorkirov: I'm not sure the extra flexibility would ever be used and it makes the component a lot easier to misuse. 

e.g. if some one does this:
`useEffect(() => { subscribeToResetUserFlow('/confirm-brc20-tx') }, []);`

instead of this:
`useEffect(() => subscribeToResetUserFlow('/confirm-brc20-tx'), []);`

Then we're opening up to a potential memory leak ( a very slow one, but it's there none the less).

But if we were to move everything into the component, we would just have to do:
`subscribeToResetUserFlow('/confirm-brc20-tx')`

and the component would clean up after itself when it's unloaded.

src/app/screens/confirmBrc20Transaction/index.tsx
victorkirov: After sleeping on this, I actually think it might be better to use the hook here 🙈 

Firstly, it would clean up this chunk of code into a single line of code at the cost of 1 fee calc which we would do on a fee rate change anyways (not super expensive).

Secondly, it would compartmentalise this component to calculate its own state instead of relying on an external input to tell it what the initial fee would be, removing the potential for a bug coming in from an external component.

What do you think?

src/app/screens/confirmBrc20Transaction/editFees.tsx
teebszet: it renders two svg rectangles

src/app/screens/sendBrc20/index.tsx
teebszet: I meant like this?

```suggestion
    if (!resultRegex.test(newValue)) {
      setAmountToSend('');
    } else {
      validateAmount(newValue);
      setAmountToSend(newValue);
    }
    const newValue = Number(e.target.value);
    if (!Number.isFinite(newValue)) {
      validateAmount(newValue);
      setAmountToSend(newValue);
    }
```

for example Number('Infinity') returns Infinity:
image

src/app/screens/confirmBrc20Transaction/index.tsx
teebszet: ok I can get behind that. the broadcast and close function would be moved to its own hook in that case

src/app/screens/confirmBrc20Transaction/index.tsx
teebszet: IMO the best way would be to expose the estimateFee function from the hook instead of calling it inside a useEffect, in the same way that the execute hook is structured. then we get the code encapsulation on both send and confirm screens, as well as the option to refetch on first render of confirm.

I don't have a huge problem with refetching on first render of confirm screen, except that this pattern of passing state around with location.state is the established pattern throughout the app. IMO that shouldn't be the pattern, and we could be using react context API for flows which couple screens together like this, but that's a new pattern that I would want to read more code before I tackle.

victorkirov: Yeah, that's a really good idea 👍 I think the state stuff makes sense in React Native but not really for web.

For the hook, what do you think about adding a boolean arg called `skipInitialFetch` which defaults to false? Then you can set it to true here and get the desired outcome.

teebszet: sounds good 👍

src/app/screens/sendBrc20/index.tsx
teebszet: It would have the opposite problem if I store it as number, where I'd have to cast it to String to set the input value.

I usually like to keep the state for input boxes as string, to match the html input value field.
pros:
- default unfilled input boxes can have value="", whereas value="0" is often not a good default
- if you cast default 0 back to string, you get a truthy value, which makes it more verbose trying to validate whether the input was filled or not: 
image
- casting to number is opt-in, and occurs just in time (not prematurely) for, e.g. validation rules, or passing on to function params.

victorkirov: Ah, good point 👍 I never thought about the cast back to string before.

src/app/screens/sendBrc20/index.tsx
victorkirov: Cool, that makes sense 👍

src/app/screens/confirmBrc20Transaction/index.tsx
victorkirov: Cool, will add it now

victorkirov: Ok, it's in the latest version 👍

src/app/screens/confirmBrc20Transaction/index.tsx
victorkirov: Yeah, something like this:
```
export const useResetUserFlow = (path: UserFlowConfigKey) => {
  const navigate = useNavigate();

  useEffect(() => {
    const broadcastChannel = new BroadcastChannel(resetUserFlowChannel);
    broadcastChannel.onmessage = (message) => {
      if (message.data !== 'reset') {
        return;
      }

      // TODO: infer UserFlowConfigKey from location?
      if (!userFlowConfig[path]) {
        return;
      }
      navigate(userFlowConfig[path]?.resetTo);
    };

    return () => {
      broadcastChannel.close();
    };
  }, []);
};

export const broadcastResetUserFlow = () => {
  const broadcastChannel = new BroadcastChannel(resetUserFlowChannel);
  broadcastChannel.postMessage('reset');
  broadcastChannel.close();
};
```

Just note that this will result in the popup broadcasting the message would also receive it in this case, which I don't think is major.

teebszet: indeed it simplified the usage a lot. thanks!

src/app/screens/confirmBrc20Transaction/brc20FeesComponent.tsx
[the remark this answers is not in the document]
teebszet: I committed a change to editFees which reverts the change on click cancel. it results in a flicker of the estimated fees when the user closes editFees modal, but I think it's OK.

otherwise i'd have to refactor to do more state management in the parent component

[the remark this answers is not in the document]
victorkirov: Cool, that's a better solution 👍 Just a note to test what happens if you set the fee rate to like 20 and then click on cancel. Is the confirm button disabled until the page settles with the new fees or can you cancel the fee change and immediately click on confirm? The former is preferable if possible.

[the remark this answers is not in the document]
teebszet: can disable the cancel and confirm buttons while fee is loading 👍

src/app/screens/confirmBrc20Transaction/index.tsx
teebszet: let's address this in another PR, when design is finalised for: https://linear.app/xverseapp/issue/ENG-2785/notice-for-an-non-taproot-address-when-sending-brc-20-tokens

src/app/utils/brc20.ts
teebszet: i'll rename this in another PR, when design is finalised for: https://linear.app/xverseapp/issue/ENG-2785/notice-for-an-non-taproot-address-when-sending-brc-20-tokens

victorkirov: 👍

src/app/screens/confirmBrc20Transaction/index.tsx
victorkirov: Shweet 👍

victorkirov (dismissed):
Noice! LGTM! 😄

victorkirov (dismissed):

src/app/routes/index.tsx
```diff
@@ -182,14 +182,6 @@ const router = createHashRouter([
...
         ),
       },
-      // {
-      //   path: 'review-ledger-stx-tx',
-      //   element: <ReviewLedgerStxTransaction />,
-      // },
-      // {
-      //   path: 'review-ledger-ft-tx',
-      //   element: <ReviewLedgerFtTransaction />,
-      // },
```
victorkirov: 🤔 Are these commented out on purpose?

teebszet: they were relics of pre- stx ledger support, that got removed in another PR which has now been merged to develop

victorkirov: 😅

DuskaT021:
@teebszet this branch has conflicts

DuskaT021:
@teebszet left comments in the [ENG-2393](https://linear.app/xverseapp/issue/ENG-2393/1-step-brc-20-transfers-web-extension)

DuskaT021:
this can be merged 👍 @teebszet

victorkirov (approved):
LGTM 🚢

PR #587, opened 2023-08-31
fix: allow display of brc-20 if application/json type

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
brc-20 inscriptions can be text/plain or applicaiton/json type. We only support text/plain at the moment.

brc-20 coin tickers (names) are also case agnostic but we generate different colour backgrounds if the case is different. 

Issue Link: ENG-2767

# ⚠️ Note
The fix makes all tickers of the same type the same colour, regardless of case, but it doesn't change the case of the ticker that was used when displaying the preview. Something to consider is that we could change all to lower/upper case for consistency.

# 🔄 Changes
- Added support for application/json BRC-20 inscriptions.
- Made colours consistent between different case BRC-20s.

# 🖼 Screenshot / 📹 Video
Before:
image

After:
image

# ✅ Review checklist


teebszet (approved):

DuskaT021:
this can be merged @teebszet

PR #589, opened 2023-09-05
fix: don't show loader on balance refetch on home screen ENG-2757

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
When the wallet is on the home screen, it often refetches balances and we display a loader. It also happens if you click off the popup and back onto it. It is a bit annoying and a lot of users have complained about it.

# 🔄 Changes
This changes how we calculate when to show the loader to only show it on the initial page load and not on a refetch.

# 🖼 Screenshot / 📹 Video
Before:

https://github.com/secretkeylabs/xverse-web-extension/assets/3322667/33817625-a2fa-4d8c-a2d3-0ea03f112ca0

After:

https://github.com/secretkeylabs/xverse-web-extension/assets/3322667/ef274be2-cf75-418c-8dbd-74e876d1c99e

# ✅ Review checklist

### DIFF
--- a/src/app/screens/home/index.tsx
+++ b/src/app/screens/home/index.tsx
@@ -1,5 +1,4 @@
 /* eslint-disable no-await-in-loop */
-import styled from 'styled-components';
 import SIP10Icon from '@assets/img/dashboard/SIP10.svg';
 import ArrowDownLeft from '@assets/img/dashboard/arrow_down_left.svg';
 import ArrowUpRight from '@assets/img/dashboard/arrow_up_right.svg';
@@ -7,14 +6,15 @@ import BitcoinIcon from '@assets/img/dashboard/bitcoin_icon.svg';
 import BitcoinToken from '@assets/img/dashboard/bitcoin_token.svg';
 import CreditCard from '@assets/img/dashboard/credit_card.svg';
 import ListDashes from '@assets/img/dashboard/list_dashes.svg';
-import { Plus } from '@phosphor-icons/react';
 import ordinalsIcon from '@assets/img/dashboard/ordinalBRC20.svg';
-import Swap from '@assets/img/dashboard/swap.svg';
 import stacksIcon from '@assets/img/dashboard/stack_icon.svg';
+import Swap from '@assets/img/dashboard/swap.svg';
 import AccountHeaderComponent from '@components/accountHeader';
 import BottomModal from '@components/bottomModal';
+import ActionButton from '@components/button';
 import ReceiveCardComponent from '@components/receiveCardComponent';
-import { isLedgerAccount } from '@utils/helper';
+import ShowBtcReceiveAlert from '@components/showBtcReceiveAlert';
+import ShowOrdinalReceiveAlert from '@components/showOrdinalReceiveAlert';
 import BottomBar from '@components/tabBar';
 import TokenTile from '@components/tokenTile';
 import useAppConfig from '@hooks/queries/useAppConfig';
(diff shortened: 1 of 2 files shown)


DuskaT021:
@victorkirov branch needs updating

src/app/screens/home/index.tsx
```diff
@@ -163,13 +163,43 @@ function Home() {
...
+    loadingStxWalletData,
+    refetchingStxWalletData,
+    loadingBtcWalletData,
+    refetchingBtcWalletData,
+    loadingCoinData,
+    refetchingCoinData,
+    loadingBtcCoinData,
+    refetchingBtcCoinData,
+  ]);
+
```
teebszet, line 202: what if we just stop displaying loading states on isRefetching? that would stop the annoying loading states on refocus.

I'm also not sure we need to trigger refetches on every initial render of home dashboard, but if we do, we could do something like query invalidations instead?

not sure if you've seen these already, but react-query has utils like: 

refetchOnWindowFocus: https://tanstack.com/query/v4/docs/react/guides/window-focus-refetching
query invalidation: https://tanstack.com/query/v4/docs/react/guides/query-invalidation
mutation invalidation: https://tanstack.com/query/v4/docs/react/guides/invalidations-from-mutations

victorkirov: I initially stopped displaying on refetch, but if the query lib has a value in cache, the loading status never seems to be true again. It always refetches. This means that if I close the popup and then open it again in a few minutes/hours/days during which time I've received funds, the popup initially shows the old balance and then updates to the correct one once the refetch is done.

I didn't consider the TTL of the cache though. We could lower it and I think it will use the loading state in that case. Will play around a bit 👍

blehitsahmad:
@victorkirov Is this ready for QA?

victorkirov:
> @victorkirov Is this ready for QA?

Nope, I'll need to address the feedback from Tim. Will do it tomorrow 👍

victorkirov:
@teebszet can you please check the last commit for an alternative approach? Let me know which you prefer.

Alternatively, we can just live with the balance never reloading and if the user has been offline for a few days and opens their wallet, they'll initially get their old balance before the refetch is done in the background and they're shown their new balance.

src/app/screens/home/index.tsx
```diff
@@ -464,7 +433,7 @@ function Home() {
                   <TokenTile
                     title={coin.name}
                     currency="FT"
-                    loading={displayLoader.coin}
+                    loading={loadingCoinData}
```
teebszet, line 436: agree these probably don't need refetching UI state

src/app/hooks/queries/useBtcWalletData.ts
```diff
@@ -1,26 +1,37 @@
...
     dispatch(SetBtcWalletDataAction(btcBalance));
     return btcData;
   };
 
+  if (clearOnLoad && !loaded.current) {
+    queryClient.removeQueries({ queryKey: [`wallet-data-${btcAddress}`], exact: true });
+  }
+
+  loaded.current = true;
+
```
teebszet, line 31: nice

teebszet (approved):

src/app/screens/home/index.tsx
```diff
@@ -417,6 +384,8 @@ function Home() {
 
   return (
     <>
+      {loadingBtcWalletData && <div>LOADiNG....</div>}
+      {isRefetching && <div>REFETCH....</div>}
```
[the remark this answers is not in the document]
victorkirov: yup 😋

victorkirov:
@teebszet I sent Quentin all the options and he approved this one 👀

teebszet (approved):
LG

DuskaT021:
this can be merged 👍

PR #594, opened 2023-09-07
fix: don't show fees until we have UTXOs on inscription screen

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
We show fees before we have UTXOs in the inscription screen resulting in a flash when we actually get the UTXOs and recalculate.

# 🔄 Changes
Keeps the state as loading until the UTXOs are pulled.

# ✅ Review checklist


yknl (approved):
Works well!

PR #602, opened 2023-09-15
feat: notice for an non taproot address when sending brc 20 tokens

teebszet:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background

Issue Link: https://linear.app/xverseapp/issue/ENG-2797/notice-for-an-non-taproot-address-when-sending-brc-20-tokens-web

requires: https://github.com/secretkeylabs/xverse-core/pull/241

# 🔄 Changes
send brc20 one step screen:
- add callout "Make sure the recipient's wallet supports BRC-20 tokens"
- add input feedback "You are transferring to yourself" when address is selected account's ordinals address or bitcoin address
- change input feedback for errors to revamped colors with icons

confirm brc20 one step screen:
- add callout if recipient is non-taproot address: "Make sure the recipient's wallet supports BRC-20 tokens"
- add callout if recipient is selected account's bitcoin or ordinals address: "You are transferring to yourself"

Impact:
- send brc20 one step screen now overflows when both inputs have feedback. ensure scrolling is not broken
- confirm brc20 one step screen changes

# 🖼 Screenshot / 📹 Video
full screen mode and popup mode (note the scrolling):
image

confirm screen:
image

# ✅ Review checklist


src/app/ui-library/inputFeedback.tsx
```diff
@@ -0,0 +1,50 @@
...
+const colors = {
+  info: Theme.colors.white_200,
+  danger: Theme.colors.danger_light,
+};
+const getColorForVariant = (variant: FeedbackVariant) => colors[variant];
+
+const icons = {
+  info: Info,
+  warning: WarningCircle,
+  success: CheckCircle,
```
victorkirov, line 18: Do we need these if the variants on feedback don't support them?

teebszet: good catch, thanks. initially I had 4 variants and then decided to not preempt design 😝

src/app/screens/sendBrc20OneStep/brc20TransferForm.tsx
```diff
@@ -193,10 +196,9 @@ function Brc20TransferForm(props: Props) {
...
             </InputFieldContainer>
           </AmountInputContainer>
-          <ErrorContainer>
-            <ErrorText>{recipientError}</ErrorText>
-          </ErrorContainer>
+          <ErrorContainer>{recipientError && <InputFeedback {...recipientError} />}</ErrorContainer>
         </InputGroup>
+        <StyledCallout bodyText={t('MAKE_SURE_THE_RECIPIENT')} />
       </div>
       <NextButtonContainer>
```
victorkirov, line 203: From the screenshot, something looks funky with the next button container. Should it not have a margin at the top above the button?

src/app/ui-library/callout.tsx
```diff
@@ -0,0 +1,94 @@
...
+export function Callout({
+  titleText,
+  bodyText,
+  variant = 'info',
+  redirectText,
+  onClickRedirect = () => {},
+}: CalloutProps) {
+  const icon = getIconForVariant(variant);
+  return (
+    <Container variant={variant}>
```
victorkirov, line 73: Aren't you getting errors in the console for these? Something about 'are you sure this is supposed to be a prop'? I think styled components need a $ infront of the var name
https://styled-components.com/docs/basics#passed-props

src/app/screens/sendBrc20OneStep/index.tsx
```diff
@@ -29,16 +30,19 @@ function SendBrc20Screen() {
...
+  const [recipientError, setRecipientError] = useState<InputFeedbackProps | null>(null);
   const [recipientAddress, setRecipientAddress] = useState('');
   const [processing, setProcessing] = useState(false);
 
   useResetUserFlow('/send-brc20');
 
   const isNextEnabled =
-    !amountError && !recipientError && !!recipientAddress && amountToSend !== '';
+    !isDangerFeedback(amountError) &&
+    !isDangerFeedback(recipientError) &&
```
victorkirov, line 43: isDangerFeedback is not imported in this file

src/app/ui-library/callout.tsx
```diff
@@ -0,0 +1,94 @@
...
+  border-radius: ${(props) => props.theme.radius(2)}px;
+  align-items: flex-start;
+  background-color: ${(props) => getBackgroundForVariant(props.variant)};
+  padding: ${(props) => props.theme.spacing(8)}px;
+  gap: ${(props) => props.theme.spacing(8)}px;
+`;
+
+const TextContainer = styled.div`
+  display: flex;
+  flex-direction: column;
```
victorkirov, line 42: I don't think this does anything 👀 This would just behave the same way as a div with no display set, no?

src/app/screens/sendBrc20OneStep/brc20TransferForm.tsx
victorkirov: Also, it's transparent 🙈 

image

src/app/screens/confirmBrc20Transaction/index.tsx
```diff
@@ -269,34 +301,25 @@ export function ConfirmBrc20Transaction() {
...
-            ))}
+            </RecipientCardContainer>
             <TransactionDetailComponent
               title={t('CONFIRM_TRANSACTION.NETWORK')}
               value={network.type}
             />
-            <TransactionDetailComponent
-              title={t('CONFIRM_BRC20.BITCOIN_VALUE')}
-              value={
-                <NumericFormat
```
victorkirov, line 292: The import for NumericFormat should be removed

src/app/ui-library/callout.tsx
victorkirov: Actually, the current styling results in this off-center behaviour:
image

You can fix it by keeping these and adding `justify-content: center;` here and changing to `align-items: stretch;` on line 34

src/app/ui-library/callout.tsx
teebszet: I don't see these console errors.. do you?

maybe it's only necessary to include $ if the prop name matches a valid attribute name?

src/app/ui-library/callout.tsx
teebszet: fixed up styling when titleText and redirectText are both also provided:
 
image

src/app/screens/sendBrc20OneStep/brc20TransferForm.tsx
teebszet: nice one. fixed here:

image

victorkirov (approved):
LGTM! 🚀 Could you just check that callout alignment is centered as per the comment?

victorkirov (changes requested):
image

I tested it out and it's still not centered 😅

src/app/ui-library/callout.tsx
victorkirov: Could you try out the changes I noted? It should center the text nicely when it's just a single line of the body with no title and link.

teebszet:
I don't think it's meant to be centered?
image

src/app/ui-library/callout.tsx
teebszet: you mean vertically centered right? not sure on what the expected behaviour is for single line actually, but zeroheight is all top aligned: 

image

teebszet: but also the figma has this: 
image

teebszet: one line: 
image

multi-line: 
image

src/app/ui-library/callout.tsx
victorkirov: Interesting. Maybe it was only introduced in version 6 🤔 Will mark as resolved 👍

src/app/ui-library/callout.tsx
```diff
@@ -0,0 +1,108 @@
...
+const getIconForVariant = (variant: CalloutVariant) =>
+  createElement(icons[variant], {
+    weight: 'fill',
+    size: '24',
+    color: Theme.colors.white_200,
+    style: {
+      flexShrink: 0,
+      flexGrow: 0,
+    },
+  });
```
victorkirov, line 32: nite: Would prefer to consistently use `styled` and keep away from lower level react functions

```suggestion
const getIconForVariant = (variant: CalloutVariant) => styled(icons[variant])`
  flex-shrink: 0;
  flex-grow: 0;
  color: ${Theme.colors.white_200};
`;
```

and in the render:
```
  const Icon = getIconForVariant(variant);
  return (
    <Container variant={variant}>
      <Icon size={24} weight="fill" />
```

victorkirov (dismissed):
LGTM 👍  Just added 1 nit, but it's not necessary to change if you disagree 🙂

src/app/ui-library/callout.tsx
teebszet: yeah true. I think styled-components has a way to pass attributes, which was what I was originally using createElement for. I'll look into this

victorkirov: You can use the code in my comment above. It's tested and working 😋

teebszet: get some weird render behaviour and this warning with that approach. needs something different: 
image

victorkirov (approved):

src/app/ui-library/callout.tsx
victorkirov: Ah, ok. I guess it creates a new one every time 😅

victorkirov: This works:
```
const getStyledIconForVariant = (icon: Icon) =>
  styled(icon)`
    color: ${(props) => props.theme.colors.white_200},
    flex-shrink: 0,
    flex-grow: 0,
`;

const icons = {
  info: getStyledIconForVariant(Info),
  warning: getStyledIconForVariant(WarningCircle),
  success: getStyledIconForVariant(CheckCircle),
  danger: getStyledIconForVariant(Warning),
};
```

and in render:
```
const VariantIcon = icons[variant];
  return (
    <Container variant={variant}>
      <VariantIcon size={24} weight="fill" />
      <TextContainer>
```

victorkirov (dismissed):

victorkirov (dismissed):

victorkirov (approved):

DuskaT021:
this can be merged 👍

PR #605, opened 2023-09-15
chore: add import sort config and put aliases into groups

fedeerbes:
# 🔘 PR Type

- [ ] Bugfix
- [ ] Enhancement
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
This PR adds import sorting config to eslint and the vscode setting to fix eslint issues on save

# 🔄 Changes
- Add vscode setting to workspace
- Add import sorting config to eslint

# 🖼 Screenshot / 📹 Video
| BEFORE | AFTER |
| :---: | :---: |
|  |  |

# ✅ Review checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -40,7 +40,64 @@
         "argsIgnorePattern": "^_"
       }
     ],
-    "no-plusplus": "off"
+    "no-plusplus": "off",
+    "import/newline-after-import": ["error", { "count": 1 }],
+    "import/order": [
+      "error",
+      {
+        "groups": ["external", "internal", "builtin", "parent", "sibling", "index"],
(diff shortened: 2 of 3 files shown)


teebszet (dismissed):

teebszet:
@fedeerbes does it rely on vscode to fix the import sort order on save?
I wonder if there's a solution which can include it with prettier on save, since eslint fix could introduce some unwanted behaviour changes

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
```
teebszet, line 3: is this going to potentially introduce code behaviour changes, if a dev doesn't carefully check what they are committing?

prettier on save is safer, because usually it doesn't modify the compiled JS

fedeerbes:
@teebszet moving this to draft until we agree on expected behavior

teebszet:
@fedeerbes I think we should use this one instead: https://www.npmjs.com/package/prettier-plugin-organize-imports

as it produces the same output as vscode's typescript language service flag for `"source.organizeImports": true`

PR #621, opened 2023-10-25
feat: allow unpublished UTXOs as inputs to PSBTs

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
We received a bug report form a market place that their PSBT signing stopped working. They narrowed it down to the bundle call failing with a 400/404 when the app tried to check if there were any inscribed/rare sats in the UTXO and the UTXO didn't exist yet as it was an output of an unpublished txn.

# 🔄 Changes
Check for a 404 response from the backend for non-existent UTXOs and ignore it when checking for ordinals in a PSBT signing request UTXO.

# Testing
To test, run this branch alongside the sats-connect example app. On the sign transaction component, go to the utils file with the createPSBT function and change the txid of the UTXO from the payment address to a random, non-existent tx id (line 71).

# ✅ Review checklist

### DIFF
--- a/src/app/hooks/useDetectOrdinalInSignPsbt.ts
+++ b/src/app/hooks/useDetectOrdinalInSignPsbt.ts
@@ -1,5 +1,6 @@
 import { getUtxoOrdinalBundle, ParsedPSBT } from '@secretkeylabs/xverse-core';
 import { BundleItem, mapRareSatsAPIResponseToRareSats } from '@utils/rareSats';
+import { isAxiosError } from 'axios';
 import { useEffect, useState } from 'react';
 import useWalletSelector from './useWalletSelector';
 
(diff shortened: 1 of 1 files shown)


src/app/hooks/useDetectOrdinalInSignPsbt.ts
```diff
@@ -15,21 +16,30 @@ const useDetectOrdinalInSignPsbt = (parsedPsbt: '' | ParsedPSBT) => {
...
+              if (item.type === 'unknown') {
+                return;
+              }
+              bundleItems.push(item);
+            });
+          } catch (e) {
+            // we get back a 404 if the UTXO is not found, so it is likely this is a UTXO from an unpublished txn
+            if (!isAxiosError(e) || e.status !== 404) {
+              // rethrow error if response was not 404
+              throw e;
```
teebszet, line 34: so this maps to a rejected promise?

teebszet: I guess it doesn't matter since we don't use the result of the Promise.all

victorkirov: Yeah. If this happens then the function will returned a rejected promise and we'd technically go back to the current behaviour with the endless loading indicator. 

Maybe we should make a ticket to show an error stating that we are having difficulty confirming the Ordinals/rare sats in the UTXOs so it's risky to publish the txn in this scenario.

teebszet (approved):

teebszet:
actually @DuskaT021 this will be a bit hard to validate, because @victorkirov also fixed it from the API side.

PR #653, opened 2023-11-10
feat: revamp stacks nfts send to screen

teebszet:
# 🔘 PR Type
- [x] Bugfix
- [x] Enhancement

# 📜 Background
https://linear.app/xverseapp/issue/ENG-3188/send-stx-nft-screen-in-fullscreen

# 🔄 Changes
- fix: ledger accounts should only open in new tab when not already in full screen
- fix: back button display should rely on back history not ledger account
- feat: revamp the send nfts screen to match send ordinals layout (responsive)

Impact:
- send nft screen and send ordinal were refactored to share the same layout, so can check for layout breakages and functional breakages on both screens (extension and full screen view)
- any flow which can open a new tab while using ledger account, now only opens the new tab if not already in full screen
  - any Send button
  - any Add Stacks Address button
  - recover assets button from settings page
- both send nft and send ordinal screens were previously hiding back button if ledger account. but if the user did 'open gallery' while on the collectibles tabs page, and then clicked through to 'send', we should allow the user to go back

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/ea522391-a1cf-4b5e-a1a6-3fc38cfeca1d

# ✅ Review checklist

### DIFF
--- a/src/app/hooks/queries/useBnsName.ts
+++ b/src/app/hooks/queries/useBnsName.ts
@@ -1,10 +1,10 @@
+import { fetchAddressOfBnsName, getBnsName, validateStxAddress } from '@secretkeylabs/xverse-core';
 import { useEffect, useState } from 'react';
-import { StacksNetwork, validateStxAddress } from '@secretkeylabs/xverse-core';
-import { fetchAddressOfBnsName, getBnsName } from '@secretkeylabs/xverse-core/api';
-import useWalletSelector from '../useWalletSelector';
 import useNetworkSelector from '../useNetwork';
+import useWalletSelector from '../useWalletSelector';
 
-export const useBnsName = (walletAddress: string, network: StacksNetwork) => {
+export const useBnsName = (walletAddress: string) => {
+  const network = useNetworkSelector();
   const [bnsName, setBnsName] = useState('');
 
   useEffect(() => {
@@ -12,15 +12,15 @@ export const useBnsName = (walletAddress: string, network: StacksNetwork) => {
       const name = await getBnsName(walletAddress, network);
       setBnsName(name ?? '');
     })();
-  }, [walletAddress]);
+  }, [walletAddress, network]);
 
   return bnsName;
 };
 
-export const useBNSResolver = (
+export const useBnsResolver = (
   recipientAddress: string,
   walletAddress: string,
-  currencyType: string,
+  currencyType?: string,
 ) => {
   const { network } = useWalletSelector();
   const selectedNetwork = useNetworkSelector();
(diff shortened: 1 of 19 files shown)


src/app/hooks/queries/useBnsName.ts
```diff
@@ -43,7 +43,7 @@ export const useBNSResolver = (
         setAssociatedAddress('');
       }
     })();
-  }, [recipientAddress]);
+  }, [recipientAddress, network, currencyType, selectedNetwork, walletAddress]);
```
fedeerbes, line 46: what is the difference between network and selectedNetwork?

src/app/layouts/sendLayout.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <ScrollContainer>
+        <Container>
+          {isScreenLargerThanXs && !hideBackButton && onClickBack && (
+            <Button onClick={onClickBack}>
+              <ArrowLeft size={20} color="white" />
+            </Button>
+          )}
+          {children}
+        </Container>
+        {isScreenLargerThanXs && (
```
fedeerbes, line 86: it will be cool to have a component which handles rendering according to break points. something like 

```ts
<RenderBPLayout
    xs={<Component/>}
    m={ <Component/>}
    />
```

or 

```ts
<RenderBPLayout
    >
      {(breakpoint) => {
        switch (breakpoint) {
          case 'xs':
            return <Component/>;
          case 'm':
            return <Component/>;
          default:
            return <Component/>;
      }
      }
    </RenderBPLayout>
```

src/app/screens/sendNft/index.tsx
```diff
@@ -18,93 +15,93 @@ import {
...
+  color: props.theme.colors.white_200,
   display: 'flex',
-  flexDirection: 'row',
-  marginLeft: '15%',
-  marginTop: props.theme.spacing(40),
+  flex: 1,
 }));
 
-const Button = styled.button((props) => ({
+const AmountInputContainer = styled.div<{ error: boolean }>((props) => ({
```
fedeerbes, line 58: we should consider making components from this to avoid code repetition, right?

teebszet: yeah true. could be the right time to make the input component in Ui library

src/app/hooks/queries/useBnsName.ts
teebszet: One is NetworkSettings, the other is StacksNetwork

PR #666, opened 2023-11-17
refactor: collectibles fetch

teebszet:
# 🔘 PR Type
- [x] Refactoring (no functional changes, no API changes)
- [x] Bug Fix

# 📜 Background
issue: https://linear.app/xverseapp/issue/ENG-3257/adapt-extension-to-fetch-nft-details

# 🔄 Changes
- refactor: use new stx fetch logic (decouple collections fetch from image+details fetch)
- feat: virtualise the grid on nft dashboard and nft collection screen to load image+details just in time

Impact:
- nfts tab
- nft collection screen
- nft detail screen
- nft send screen
- nft confirm tx screen

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/5544aeb4-349a-40cb-8f5a-d2740cdd29b1

# ✅ Review checklist

### DIFF
--- a/src/app/hooks/queries/useStacksCollectibles.ts
+++ b/src/app/hooks/queries/useStacksCollectibles.ts
@@ -1,8 +1,8 @@
 import useNetworkSelector from '@hooks/useNetwork';
 import useWalletSelector from '@hooks/useWalletSelector';
 import { getNftCollections, StacksCollectionList } from '@secretkeylabs/xverse-core';
-import { useInfiniteQuery } from '@tanstack/react-query';
-import { handleRetries, InvalidParamsError } from '@utils/query';
+import { useQuery } from '@tanstack/react-query';
+import { handleRetries } from '@utils/query';
 
 const useStacksCollectibles = () => {
   let { stxAddress } = useWalletSelector();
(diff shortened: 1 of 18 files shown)


src/app/hooks/queries/useStacksCollectibles.ts
```diff
@@ -14,28 +14,15 @@ const useStacksCollectibles = () => {
...
-      if (currentLength < lastpage.total) {
-        return currentLength;
-      }
-      return false;
-    },
-    refetchOnMount: false,
-    refetchOnWindowFocus: false,
+    queryKey: ['nft-collection-data', stxAddress],
+    queryFn: fetchNftCollections,
+    staleTime: 5 * 60 * 1000, // 5mins
```
fedeerbes, line 25: question: do we want to wait that long here. since the new implementation is light weight than previous one? and probably won't hit the rate limit?

teebszet: Maybe 1min is better?

src/app/screens/nftCollection/index.tsx
```diff
@@ -110,9 +115,59 @@ const StyledGridContainer = styled(GridContainer)`
...
+function CollectionGridItemWithData({
+  nft,
+  collectionData,
+  isGalleryOpen,
+}: {
+  nft: NonFungibleToken;
+  collectionData: StacksCollectionData;
+  isGalleryOpen: boolean;
+}) {
+  const { data: nftData } = useNftDetail(nft.identifier);
```
fedeerbes, line 144: if i'm not wrong the API call will fail when is bns contract. should we prevent the call?

fedeerbes (approved):

src/app/hooks/queries/useStacksCollectibles.ts
fedeerbes: i was thinking about two but since extensions doesn't have a refetch method one may be better, wdyt?

PR #669, opened 2023-11-21
fix: fiat rates not fetching when user changed currency

teebszet:
# 🔘 PR Type
- [x] Bugfix

# 📜 Background
https://secretkeylabs.slack.com/archives/C05QCJM9D1N/p1700550855171529

# 🔄 Changes
- fix: use correct cache keys where staleTime was introduced, but previously cache keys were too broad

Impact:
- also made sure other changed queries in the same commit have enough params in their query cache keys (it could have resulted in bugs where data was cached between switching testnet/mainnet, but would have been very rare because you the other queries were still caching by id as well)

# 🖼 Screenshot / 📹 Video

https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/f15a077c-87e1-4e43-aafa-385235eef220

# ✅ Review checklist


jordankzf (approved):
Tested, works well. The fiat amount now updates instantly.

DuskaT021:
tested we can merge @teebszet

PR #710, opened 2023-12-13
Fix account referencing issue on login

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
On login we were using the account id as an index of the account list, which broke when the deletion of Ledger accounts was added.

# 🔄 Changes
We now reference the accounts by ID.

To reproduce on develop branch:
- Create 2 ledger accounts
- Delete the first one
- Switch to the one that's left
- Lock your wallet
- Unlock with password
- Wallet should be a bit broken at this stage
- Lock your wallet
- Close the popup and open again and you should be met with the onboarding screen

This should work fine on this branch

Should also test:
- onboarding flow
- login flow

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist

### DIFF
--- a/src/app/components/guards/auth.tsx
+++ b/src/app/components/guards/auth.tsx
@@ -1,13 +1,13 @@
-import { PropsWithChildren, useEffect } from 'react';
-import { useNavigate } from 'react-router-dom';
 import useSeedVault from '@hooks/useSeedVault';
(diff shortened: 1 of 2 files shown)


dhriaznov (approved):
Thanks for fixing this @victorkirov 🙌

PR #720, opened 2023-12-18
[ENG-3471] feat: use estimated fees for stx transactions

teebszet:
# 🔘 PR Type
- [x] Bugfix
- [x] Enhancement

# 📜 Background
https://linear.app/xverseapp/issue/ENG-3458

depends on https://github.com/secretkeylabs/xverse-core/pull/323 and
https://github.com/secretkeylabs/xverse-core/pull/335

# 🔄 Changes

- update to core version where we use the correct estimateTransaction function for STX transactions (sendStx, sendNft, sendFt)
- update to core version where we have applyFeeMultiplier logic in one place
- applyFeeMultiplier also caps the fee at the thresholdHighStacksFee
- use applyFeeMultiplier in sendFt, sendNft, sendStx, and transactionRequest (popup)

Impact:
- impacts sendFt, sendNft, sendStx
- also impacts dapps which trigger a transaction request popup

# 🖼 Screenshot / 📹 Video

send stx (estimates are returning high fees. screenshot shows capped at threshold: 1STX):
image

send nft:
image

send ft:
image

# ✅ Review checklist


src/app/screens/transactionRequest/index.tsx
```diff
@@ -55,11 +55,7 @@ function TransactionRequest() {
       tokenTransferPayload.amount,
       tokenTransferPayload.memo!,
```
dhriaznov, line 56: Let's also add a check for the `tokenTransferPayload.memo` here to make the code safer?

dhriaznov (commented):
Good job 👍 left a small ask below :)

src/app/screens/transactionRequest/index.tsx
teebszet: good call. it should be an optional arg in the core function

teebszet: but I don't want to refactor the core function to make it optional, because then I should also reorder the args to have memo?: string come after the required params.

maybe a compromise here is to apply the default earlier like `tokenTransferPayload.memo ?? ''`

dhriaznov (approved):

Lifei-Coder2:
why you can set a thresholdHighStacksFee as the biggest fee,  don't you consider the  success rate for the txns?

yknl:
> @teebszet hello, why you can set a thresholdHighStacksFee as the biggest fee，I wonder your reason for the highest limit value on the stx chain compared to the response of "https://api.hiro.so/v2/fees/transaction" api , don't you consider the success rate for the txns?

That is not what thresholdHighStacksFee does. It's used for high fee warnings only. And this is a very old/outdated PR.

Lifei-Coder2:
> > @teebszet hello, why you can set a thresholdHighStacksFee as the biggest fee，I wonder your reason for the highest limit value on the stx chain compared to the response of "https://api.hiro.so/v2/fees/transaction" api , don't you consider the success rate for the txns?
> 
> That is not what thresholdHighStacksFee does. It's used for high fee warnings only. And this is a very old/outdated PR.

I mean your wallet set thresholdHighStacksFee as the fee.  I  want to know how does you think of it

PR #725, opened 2023-12-20
Tim/eng 3475 nonce not updating correctly after switching accounts

teebszet:
# 🔘 PR Type
- [x] Bugfix

# 📜 Background
Issue Link: https://linear.app/xverseapp/issue/ENG-3475

there is a feature on the transaction request screen which checks a dapp tx request popup payload and does an auto account switch if there is a mismatch.

scenario 1:
if I signed in with account 1 on the dapp, and switched my extension to account 2,
expect the popup for a dapp tx will auto switch my account back to the one matching what I signed into the dapp with.

# 🔄 Changes
this fix doesn't change the behaviour, but ensures the auto switch also regenerates the unsignedTx with the correct account (and therefore displays the correct nonce in Edit Nonce)

- fix: bug where transaction request screen would initiate an auto account switch, but not regenerate the unsignedTx

Impact:
- STX transaction request screen
- only used when interacting with dapps

# 🖼 Screenshot / 📹 Video
after fix:

https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/3bbd8788-4559-4f14-9736-55e88df8702e

also tested with pending transactions, and nonce increments as expected

# ✅ Review checklist

### DIFF
--- a/src/app/screens/transactionRequest/index.tsx
+++ b/src/app/screens/transactionRequest/index.tsx
@@ -33,8 +33,7 @@ const LoaderContainer = styled.div((props) => ({
 }));
 
 function TransactionRequest() {
-  const { stxAddress, network, stxPublicKey, feeMultipliers, accountsList, selectedAccount } =
-    useWalletSelector();
+  const { network, feeMultipliers, accountsList, selectedAccount } = useWalletSelector();
   const { payload, tabId, requestToken, stacksTransaction } = useStxTransactionRequest();
   const navigate = useNavigate();
   const selectedNetwork = useNetworkSelector();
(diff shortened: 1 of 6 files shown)


m-aboelenein (approved):

teebszet:
I see some request looping. not ready for merge

src/app/screens/transactionRequest/index.tsx
```diff
@@ -146,48 +191,14 @@ function TransactionRequest() {
...
-          await handleTokenTransferRequest(payload);
-        } else if (payload.txType === 'contract_call') {
-          await handleContractCallRequest(payload);
-        } else if (payload.txType === 'smart_contract') {
-          await handleContractDeployRequest(payload);
-        }
-      } else {
-        await handleTxSigningRequest();
-      }
+      await createRequestTx(selectedAccount!);
```
dhriaznov, line 195: Do we want to add a check if `selectedAccount` is not undefined here instead of using `!` symbol to make it safer?

src/app/screens/transactionRequest/index.tsx
```diff
@@ -104,21 +108,61 @@ function TransactionRequest() {
...
+  const handleContractDeployRequest = async (
+    contractDeployPayload: ContractDeployPayload,
+    requestAccount: Account,
+  ) => {
     const response = await createDeployContractRequest(
       contractDeployPayload,
       selectedNetwork,
-      stxPublicKey,
+      requestAccount.stxPublicKey,
       feeMultipliers!,
```
dhriaznov, line 119: Same question as for `selectedAccount` below

dhriaznov (approved):
Got a couple of code suggestions, otherwise LGTM

PR #727, opened 2023-12-21
[ENG-3474] feat: Make UI for STX RBF in the extension

dhriaznov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
This feature is needed to help our Stacks users increase the chance of getting their pending transactions included in the next Stacks anchor block, by increasing the transaction fee.

Issue Link: #[[ENG-3474](https://linear.app/xverseapp/issue/ENG-3474/make-ui-for-stx-rbf-in-the-extension)]
Context Link (if applicable):

# 🔄 Changes
- Added `useRbfTransactionData` hook
- Updated the Speed Up Tx page to add the STX functionality
- Added the Speed Up button to `src/app/components/transactions/stxTransferTransaction.tsx` component
- Removed the unused code in `src/app/components/transactions/stxTransaction.tsx` component
- Removed the unused `src/assets/img/transactions/increaseFee.svg` icon

Impact:
- This PR impacts both BTC and STX RBF features

# 🖼 Screenshot / 📹 Video
The demo videos are in the linear ticket.

Monosnap Monosnap 2023-12-28 17-00-19
Monosnap Monosnap 2023-12-28 17-00-27
Monosnap Monosnap 2023-12-28 17-00-37
Monosnap Monosnap 2023-12-28 17-00-59

# ✅ Review checklist


src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,216 @@
...
+        rbfTxSummary: {
+          currentFee: fee.toNumber(),
+          currentFeeRate: fee.toNumber(),
+          minimumRbfFee: minimumFee.toNumber(),
+          minimumRbfFeeRate: minimumFee.toNumber(),
+        },
+        rbfRecommendedFees: Object.fromEntries(
+          Object.entries({
+            medium: {
+              enoughFunds: !BigNumber(mediumFee).gt(BigNumber(stxAvailableBalance)),
```
teebszet, line 132: nit: could use .lte() rather than double negative?

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,216 @@
...
+        minimumFee = microStxToStx(Math.ceil(stxToMicrostacks(minimumFee).toNumber()));
+      }
+
+      if (feeMultipliers && highFee > BigInt(feeMultipliers?.thresholdHighStacksFee)) {
+        highFee = feeMultipliers.thresholdHighStacksFee;
+      }
+
+      if (feeMultipliers && mediumFee > BigInt(feeMultipliers?.thresholdHighStacksFee)) {
+        mediumFee = feeMultipliers.thresholdHighStacksFee * 0.75;
+      }
```
teebszet, line 119: is there an edge case here, where the minimumFee can be higher than the medium/high Fees?

src/app/screens/speedUpTransaction/index.tsx
teebszet, line 1: @dhriaznov IMO we should stop writing new code with this pattern of shared screens/large components between STX/BTC 
`if(isBtc) { // do btc } else { // do stacks }`

it makes the screen file bloated, and more difficult to maintain/change in future (e.g. see ConfirmBtcTransactionComponent.tsx, or how sendForm.tsx was eventually extended to support many currencyTypes). 

If we have the time, I would much prefer to duplicate the screen, and reuse the lower level components, or reuse the layout containers

src/app/hooks/useRbfTransactionData.ts
dhriaznov: yeah, I think that there could be this edge case

src/app/screens/speedUpTransaction/index.tsx
dhriaznov: I agree, I created separate components now

src/app/hooks/useRbfTransactionData.ts
teebszet: > yeah, I think that there could be this edge case. should we disable those options then or calculate them to be higher than the current fee?

the unit tests and logic should follow the user story set out here:

https://www.notion.so/xverseapp/Speed-up-Stacks-transaction-Improve-fee-estimation-4455d27f8bdc4ebfbd94a16d7f18fe12?pvs=4#6e0e607014764f2ab9d6f8b76d677dfe

teebszet: for [SPEEDSTX-EXT-2]

teebszet: to clarify: I don't mind when you move logic into core in this case, since you're going to be working on both clients anyway.

but before we can merge, the feature should satisfy the user stories set out in notion.

since we have really clear requirements/specs now, the most efficient way would be for  the PR author to quickly double check the PR matches the notion (in case the requirements changed or no longer make sense, etc.)

thanks!

dhriaznov: totally makes sense @teebszet, thanks for noticing that! I updated the logic to follower the user story Case 2 and Case 3
1 Monosnap Monosnap 2024-01-08 18-24-35
2 Monosnap Monosnap 2024-01-08 18-25-06

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  const fetchRbfData = useCallback(async () => {
+    if (!selectedAccount || !transaction) {
+      setIsLoading(false);
+      return;
+    }
+
+    if (!isBtcTransaction(transaction)) {
+      return fetchStxData();
+    }
```
teebszet, line 232: it kinda feels like this should be two separate hooks. 

you end up loading more code than you need to, like btcClient (when doing stx rbf) and selectedNetwork, network, stxAvailableBalance (when doing btc rbf).

did you consider splitting the screens/speedUpTransaction/index.tsx into btc and stx versions? I saw you split it up on the component level instead

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+        unsignedTx.payload,
+        undefined,
+        selectedNetwork,
+      );
+
+      let minimumFee = fee.multipliedBy(1.25);
+      if (!Number.isSafeInteger(minimumFee)) {
+        // round up the fee to the nearest integer
+        minimumFee = microStxToStx(Math.ceil(stxToMicrostacks(minimumFee).toNumber()));
+      }
```
teebszet, line 157: what was the requirements for this part?

should you not do all calculations on microstacks, and only convert to stx if you need to before/in the return statement?

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+
+        feePresets = await constructRecommendedFees(
+          'higher',
+          higherFee,
+          'highest',
+          highestFee,
+          stxAvailableBalance,
+          feeMultipliers?.thresholdHighStacksFee,
+        );
+      }
```
teebszet, line 198: 👍

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+}
+
+export async function getRawTransaction(txId: string, network: SettingsNetwork): Promise<string> {
+  const baseUrl = network?.address;
+  const apiUrl = `${baseUrl}/extended/v1/tx/${txId}/raw`;
+
+  return axios.get<RawTransactionResponse>(apiUrl).then((response) => response.data.raw_tx);
+}
+
+const constructRecommendedFees = async (
```
teebszet, line 80: does this need to be async?

src/app/hooks/useRbfTransactionData.ts
dhriaznov: the requirement for the minimum fee is here https://secretkeylabs.slack.com/archives/C063ZVDLCLB/p1703778799805989?thread_ts=1703712686.356469&cid=C063ZVDLCLB

src/app/hooks/useRbfTransactionData.ts
dhriaznov: nope, it doesn't, good gatch 🙂

src/app/hooks/useRbfTransactionData.ts
dhriaznov: that's a valid point.

regarding separate hooks - my suggestion here is to see how it'll look like when we move the helper functions and some of the STX logic to the core repo and then decide.

and regarding separate screens - not sure if it's necessary since they share quite a lot of logic and UI currently. if we make the separate hooks then we probably should separate the screens as well

teebszet: > if we make the separate hooks then we probably should separate the screens as well

yeah that would make sense

> regarding separate hooks - my suggestion here is to see how it'll look like when we move the helper functions and some of the STX logic to the core repo and then decide.

OK

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+      const txRaw: string = await getRawTransaction(transaction.txid, network);
+      const unsignedTx: StacksTransaction = deserializeTransaction(txRaw);
+
+      const [slow, medium, high] = await estimateTransaction(
+        unsignedTx.payload,
+        undefined,
+        selectedNetwork,
+      );
+
+      const currentMicrostacksFee = stxToMicrostacks(fee);
```
teebszet, line 153:
```suggestion
      const currentMicrostacksFee = transaction.fee;
```

yeah the minimum fee rounding is a lot cleaner, thanks!

nit: you still have this extra arithmetic

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+  // TODO: move the STX RBF calculations to xverse-core and add unit tests
+  const fetchStxData = useCallback(async () => {
+    if (!transaction || isBtcTransaction(transaction)) {
+      return;
+    }
+
+    try {
+      setIsLoading(true);
+
+      const fee = microStxToStx(transaction.fee);
```
teebszet, line 143: const fee is not used until L203 L204?

```suggestion
```

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+          stxAvailableBalance,
+          feeMultipliers?.thresholdHighStacksFee,
+        );
+      }
+
+      setRbfData({
+        rbfTransaction: undefined,
+        rbfTxSummary: {
+          currentFee: fee.toNumber(),
+          currentFeeRate: fee.toNumber(),
```
teebszet, line 204:
```suggestion
          currentFee: microStxToStx(currentMicrostacksFee).toNumber(),
          currentFeeRate: microStxToStx(currentMicrostacksFee).toNumber(),
```

src/app/hooks/useRbfTransactionData.ts
teebszet: thanks! much cleaner

teebszet (approved):

src/app/hooks/useRbfTransactionData.ts
```diff
@@ -0,0 +1,274 @@
...
+  mempoolApi,
+  rbf,
+  RecommendedFeeResponse,
+  SettingsNetwork,
+  StacksTransaction,
+  stxToMicrostacks,
+  StxTransactionData,
+} from '@secretkeylabs/xverse-core';
+import { deserializeTransaction, estimateTransaction } from '@stacks/transactions';
+import { isLedgerAccount, microStxToStx } from '@utils/helper';
```
teebszet, line 12: nit: xverse-core already exports stxToMicrostacks and microstacksToStx
```suggestion
  stxToMicrostacks,
  microstacksToStx
  StxTransactionData,
} from '@secretkeylabs/xverse-core';
import { deserializeTransaction, estimateTransaction } from '@stacks/transactions';
import { isLedgerAccount } from '@utils/helper';
```

teebszet:
@dhriaznov thanks for make the code improvements. let's merge this after the base branch is merged

dhriaznov:
Thanks for the review @teebszet! I resolved the git conflicts and applied the above code suggestions

teebszet (dismissed):

DuskaT021:
@dhriaznov needs conflicts resolved

teebszet (approved):

PR #738, opened 2024-01-05
[ENG-3496] fix: utxo 500 issue add a fallback option to our api httpsbtc

teebszet:
# 🔘 PR Type
- [x] Enhancement

# 📜 Background
Issue Link: [ENG-3496](https://linear.app/xverseapp/issue/ENG-3496/fix-utxo-500-issue-add-a-fallback-option-to-our-api-httpsbtc)
depends on: https://github.com/secretkeylabs/xverse-core/pull/339

# 🔄 Changes
- added an (optional) input to change network screen for fallback btc api url. it can be saved with an empty input
- fixed the UX so that switching between mainnet/testnet will remember previously set urls
- renamed the Reset URL button to 'Reset to default'
- added a quick validation for BTC URL and Stacks URL inputs, to not allow empty
- uses core version https://github.com/secretkeylabs/xverse-core/pull/339 where BitcoinEsploraApiProvider will retry on the fallbackUrl (if provided) on:
  - request timeouts
  - response 4xx 5xx codes
- some minor restyling on the change network screen
  - use the revamped feedback errors
  - use the revamped input 'X' button (cross in circle icon)

Impact:
- anywhere useBtcClient was used, will have a fallback url, which is configurable from the settings screen
  - can verify this by triggering calls to btcclient and checking the baseURL of the network requests
  - btcclient is used in sendBtc, sendOrdinal, sendRareSats, restoreOrdinal, sendBrc20, btcRbf, transaction history, sign psbt request, edit btc fees, one step brc20, create inscription 
- note that BitcoinEsploraApiProvider needs to still be used without the fallback url, when validating each custom URL input (on the change network screen itself)

# 🖼 Screenshot / 📹 Video
https://github.com/secretkeylabs/xverse-web-extension/assets/6109710/16357ae5-154c-4644-87a6-244d34e483cd

# ✅ Review checklist


tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "isolatedModules": true,
     "noEmit": false,
     "jsx": "react-jsx",
-    "rootDir": "src",
+    // "rootDir": "src",
```
victorkirov, line 19: nit: delete

src/app/hooks/useBtcClient.ts
```diff
@@ -3,17 +3,18 @@ import { useMemo } from 'react';
 import useWalletSelector from './useWalletSelector';
 
 const useBtcClient = () => {
-  const { network, btcApiUrl } = useWalletSelector();
-  const { type, btcApiUrl: remoteBtcApiURL } = network;
-  const url = btcApiUrl || remoteBtcApiURL;
+  const { network } = useWalletSelector();
+  const { type, btcApiUrl, fallbackBtcApiUrl } = network;
+  const url = btcApiUrl;
```
victorkirov, line 8: Isn't this changing the behavior a bit?

My understanding is:
- The previous implementation had btcApiUrl and network.btcApiUrl in the store
- btcApiUrl was either the custom URL that the user defined in settings, or it was undefined/null if they hadn't overridden it.
- network.btcApiUrl was the default value to use if the btcApiUrl was null (not set by user)
- If a custom URL was set by the config API call, we would override that default
- If a user had set a custom URL and then reset it, we would just make btcApiUrl null and that would result in the default being used form the network.

Theoretically at least, whether it was working, I'm not sure 😅 

The problem now is that we have no distinction between a custom URL being set and not.

victorkirov: Yeah, looking at the code, I don't think the above was working 100%, but it seemed to be the intention 😅

victorkirov: To clarify the above because it was a bit of a dump of stuff, I think we need to satisfy the following test criteria:

scenario 1 - I haven't set up a custom URL:
- the default URL should be used (I think this is satisfied)
- If a new default URL comes form the config endpoint, my default URL should change (I _don't_ think this is satisfied)
- if I set a custom URL and then go back to settings and reset the URL, I should be reset to the URL from the config endpoint and further updates to the config endpoint should affect my URL (I _don't_ think this is satisfied)

scenario 2 - I have set up a custom URL
- The custom URL should be used (I think this is satisfied)
- If I go to settings and reset my URL, it should take the one from the config API endpoint and not necessarily the one from Core (I _don't_ think this is satisfied)

scenario 3 - regardless of settings
- If a new default is set from the API, that should be the default used on a URL reset. If the API config URL is removed, i.e. the endpoint returns null/undefined for the custom URL, we should go back to using the Core URL as the default

victorkirov (commented):
I left a thread of comments but maybe the idea is to not use the URL value from the config endpoint? If that's the case, then I'm happy to approve 👍

src/app/hooks/useBtcClient.ts
teebszet: Yeah that makes sense. I was wondering about what the intention was for the btcapiurl coming from app config endpoint... as it didn't seem like it was applied from the store side

teebszet: Let me revisit it tomorrow with the scenarios including updating the api url via app config endpoint

teebszet:
> I left a thread of comments but maybe the idea is to not use the URL value from the config endpoint? If that's the case, then I'm happy to approve 👍

I asked @m-aboelenein here: https://secretkeylabs.slack.com/archives/C05QCJM9D1N/p1704804634811279

and looks like we can remove that functionality

src/app/hooks/useBtcClient.ts
teebszet: @victorkirov see comment below regarding the remote btcApiUrl config 🙂

tsconfig.json
teebszet: done

src/app/hooks/useBtcClient.ts
victorkirov: Awesome!! 🎉

victorkirov (dismissed):
So much cleaner! 🎉

DuskaT021:
@teebszet needs conflicts resolved

PR #756, opened 2024-01-15
[ENG-3513] feat: View accounts balances in the Change Account screen

dhriaznov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
See --> https://www.notion.so/xverseapp/View-Accounts-balances-from-change-account-list-0b183d7d92a0446a88bc2e8bc3432f2b?pvs=4#3761a6cca9254b53bc9ef42511b61f77

Issue Link: #[[ENG-3513](https://linear.app/xverseapp/issue/ENG-3513/%F0%9F%92%B2%F0%9F%8C%8D-view-accounts-balances-in-extensions-change-account-screen)]
Context Link (if applicable):

# 🔄 Changes
- Added `useAccountBalance` hook fetching and calculating the total of BTC + STX accounts balance in a queue of requests
- Added `useIntersectionObserver` hook which helps to trigger the fetching on scroll when the item in the account list is visible on the user's screen
- Added a UI element displaying the account balance (and a loader) for each account on the Change Account screen
- When the user switches the account, we fetch the new value on the Dashboard and save it to show on the Change Account screen when the user goes back to it from the Dashboard.

Impact:
- This PR impacts the account list display on the Change Account screen

# 🖼 Screenshot / 📹 Video
account balance Monosnap Monosnap 2024-01-16 21-11-01

https://github.com/secretkeylabs/xverse-web-extension/assets/36603049/001fab65-5b5a-4637-92bc-4d4eca99c043

# ✅ Review checklist


teebszet:
reassigned the review to @fedeerbes since he will be working on the same feature for mobile

src/app/components/accountRow/index.tsx
```diff
@@ -199,6 +205,14 @@ function AccountRow({
   >();
   const { removeLedgerAccount, renameAccount, updateLedgerAccounts } = useWalletReducer();
 
+  const totalBalance = accountBalances[account?.btcAddress ?? ''];
+
+  useEffect(() => {
+    if (fetchBalance && shouldFetch && !totalBalance) {
+      fetchBalance(account);
+    }
+  }, [shouldFetch, totalBalance]);
```
fedeerbes, line 214: how about we move this logic to the parent component? we are making a decision in the child that the parents can make.

src/app/components/accountRow/index.tsx
```diff
@@ -315,6 +329,21 @@ function AccountRow({
...
+                  thousandSeparator
+                  renderText={(value: string) => <Balance isSelected={isSelected}>{value}</Balance>}
+                />
+              )}
+              {isAccountListView && !totalBalance && (
+                <Balance isSelected={isSelected}>
+                  {EMPTY_LABEL}
+                  <MoonLoader color="white" size={12} />
+                </Balance>
+              )}
```
fedeerbes, line 346: question: any reason this conditions aren't ternary?

src/app/hooks/queries/useAccountBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+    let btcBalance = 0;
+    let stxBalance = 0;
+
+    if (account.btcAddress) {
+      const btcData: BtcAddressData = await btcClient.getBalance(account?.btcAddress);
+      btcBalance = btcData.finalBalance;
+    }
+
+    if (account.stxAddress) {
+      const stxData: StxAddressData = await fetchStxAddressData(
```
fedeerbes, line 59: i think we can improve timing and resources here by avoiding the use of this method and calling directly to 

``` ts
const apiUrl = `${getNetworkURL(network)}/v2/accounts/${stxAddress}?proof=0`;
```

cause fetchStxAddressData also makes calls to tx, pending tx and process them. and we only care about balance here right?

fedeerbes (approved):
LGTM, really nice work on interceptor and queuing. left some suggestions and a question

src/app/components/accountRow/index.tsx
dhriaznov: I think it would require to make a nested ternary condition here which is usually considered not a good practice by code linters

src/app/hooks/queries/useAccountBalance.ts
dhriaznov: yeah, that definitely makes sense 👍

dhriaznov:
Thanks for the suggestions @fedeerbes , I applied them 🙌 could you take a second look please?

fedeerbes (approved):

fedeerbes:
before moving to testing we need to make sure ft token with fiat rate are part of the account balance. and maybe have a single util to calculate the balance per account 

cc @dhriaznov

dhriaznov:
I added the fungible token balance calculation @fedeerbes, could you review pls?

fedeerbes (dismissed):

dhriaznov:
I updated the PR resolving the git conflicts, adding the brc-20 balance calculation to the `calculateTotalBalance` util func and fixing the issue that Dušana found (when there were different total balance values on the Dashboard and Change Account screens), could you please review again? @fedeerbes 
Monosnap Monosnap 2024-01-24 20-03-14
Monosnap Monosnap 2024-01-24 20-03-19

dhriaznov:
I tested it with ledger account as well:

ledger Monosnap Monosnap 2024-01-24 20-10-32
ledger 2Monosnap Monosnap 2024-01-24 20-10-42

DuskaT021:
@dhriaznov fv needs branch updated

src/app/hooks/queries/useAccountBalance.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        ftCoinList = fungibleTokenList;
+      }
+    }
+
+    const totalBalance = calculateTotalBalance({
+      stxBalance,
+      btcBalance,
+      ftCoinList,
+      brcCoinsList,
```
fedeerbes, line 119: question: should this be fetched for the processing account instead of the selected one? cause it will sum the same balance to all the accounts right?

dhriaznov: yeah you're right, I think I overlooked this

src/app/hooks/queries/useAccountBalance.ts
```diff
@@ -27,6 +30,58 @@ const useAccountBalance = () => {
   const [isProcessingQueue, setIsProcessingQueue] = useState(false);
   const [queueLength, setQueueLength] = useState(0);
 
+  const fetchBrcCoinsBalances = async (ordinalsAddress: string) => {
```
fedeerbes, line 33: nit: we should have an util to handle this here and in useAccountBalance hook.

fedeerbes (approved):

src/app/hooks/queries/useAccountBalance.ts
dhriaznov: agreed 👌 I added a todo comment

fedeerbes (dismissed):

fedeerbes (approved):

PR #767, opened 2024-01-23
[ENG-3594] feature: Send Max STX

jordankzf:
# 🔘 PR Type
image
image
image
image
image
image

Notes:
1. Success screen restyle not part of this PR.
2. Confirmation screen restyle not part of this PR.
3. Estimated fees seem weird.

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
Provide a brief explanation of why this pull request is needed. Include the problem you are solving or the functionality you are adding. Reference any related issues.

Issue Link: #[issue_number]
Context Link (if applicable):

# 🔄 Changes
Enumerate the changes made in this pull request, detailing what has been modified, added, or removed. Include technical details and implications if necessary.

Impact:
- Explain the broader impact of these changes.
- How it improves performance, fixes bugs, adds functionality, etc.

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


jordankzf:
Needs testing with a Ledger device

src/app/components/inputScreen/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  // gap: ${(props) => props.theme.space.xl};
+`;
+
+type InputScreenProps = {
+  inputs: React.ReactNode;
+  buttons: React.ReactNode;
+  header?: React.ReactNode;
+};
+
+function InputScreen({ inputs, buttons, header }: InputScreenProps) {
```
victorkirov, line 27: I'm not 100% convinced this component is needed. It's currently used in 1 place with a single input and single button. Do we foresee it being used elsewhere?

src/app/components/inputScreen/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+`;
+
+const ButtonsContainer = styled.div`
+  margin: ${(props) => props.theme.spacing(12)}px 0;
+`;
+
+const InputsContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  // gap: ${(props) => props.theme.space.xl};
```
victorkirov, line 18: Comments 👀

src/app/screens/confirmStxTransaction/index.tsx
```diff
@@ -123,7 +134,7 @@ function ConfirmStxTransaction() {
     }
 
     const txAmount = new BigNumber(txPayload.amount.toString(10));
```
victorkirov, line 136: Instead of having updateUI as a method that's called from the useEffect, could you rather put it directly inside the useEffect?

Also, the useEffect will always call updateUI because on load those states will always be empty. It might be better to try and parse the values and set the initial states directly, otherwise remove the `if` in the useEffect and add an empty dependency array as the second arg.

src/app/screens/confirmStxTransaction/index.tsx
```diff
@@ -123,7 +134,7 @@ function ConfirmStxTransaction() {
...
-    const txFee = new BigNumber(unsignedTx.auth.spendingCondition.fee.toString());
+    const txFee = new BigNumber(fee.toString() ?? unsignedTx.auth.spendingCondition.fee.toString());
```
victorkirov, line 137: fee.toString() will always be non null/endefined, so it will never get the fee from the spending condition.

src/app/screens/confirmStxTransaction/index.tsx
```diff
@@ -137,7 +148,7 @@ function ConfirmStxTransaction() {
     const modifiedMemoString = txMemo.content.split('\u0000').join('');
 
     setAmount(txAmount);
-    setStateFee(txFee);
+    if (!fee) setStateFee(txFee);
```
victorkirov, line 151: This if will never fire. fee will always be an instance of BigNumber and a class instance is always truthy.

src/app/screens/sendBtc/recipientSelector.tsx
```diff
@@ -51,41 +38,44 @@ function RecipientSelector({
     setAddressIsValid(true);
   };
 
-  const inputFeedback = useMemo(() => {
+  interface InputFeedback {
+    variant: 'danger';
+    message: string;
+  }
```
victorkirov, line 44: nit: prefer types over interfaces unless we are making use of interfacer functionality. Also, this should maybe live up in the file by the other types.

src/app/screens/sendBtc/recipientSelector.tsx
```diff
@@ -1,39 +1,26 @@
...
   recipientAddress: string;
   setRecipientAddress: (address: string) => void;
   onNext: () => void;
   isLoading: boolean;
   header?: React.ReactNode;
 };
 
-// TODO: this could be extracted into a component for reuse
-function RecipientSelector({
+function BtcRecipientScreen({
```
victorkirov, line 17: nit: filename should match the component name

src/app/screens/sendBtc/recipientSelector.tsx
```diff
@@ -1,39 +1,26 @@
...
-type Props = {
+type BtcRecipientScreenProps = {
   recipientAddress: string;
   setRecipientAddress: (address: string) => void;
   onNext: () => void;
   isLoading: boolean;
   header?: React.ReactNode;
 };
 
-// TODO: this could be extracted into a component for reuse
```
victorkirov, line 29: I think this should still be extracted into a generic recipient selector component. Maybe I can do it in the base branch and you can make use of it for stacks in this PR?

src/app/ui-components/selectFeeRate/feeItem.tsx
```diff
@@ -160,8 +162,16 @@ function FeeItem({
   const mainColor = totalFee ? 'white_0' : 'white_400';
   const secondaryColor = totalFee ? 'white_200' : 'white_400';
 
+  const feesExceedBalance = Boolean(
+    totalFee && absoluteBalance && absoluteBalance < Number(totalFee),
+  );
```
victorkirov, line 167: I don't think this is correct. This would assume that you don't have enough funds if the fee is above your balance, disregarding the amount that you're sending.

e.g. If my balance is 1000 and I'm sending 950 with a fee of 100, then this will still show false because 100<1000 when I don't actually have enough balance for this fee (need 1050 but only have 1000)

src/app/ui-components/selectFeeRate/index.tsx
```diff
@@ -138,20 +142,38 @@ function SelectFeeRate({
...
+                <Label $size="s" $variant="dark">
+                  {value} {fiatUnit}
+                </Label>
+              )}
+            />
+          </RowContainer>
+        )}
       </RowContainer>
-      {fee && (
+      {fee && feeRateUnits && (
```
victorkirov, line 176: It's really hard to understand what's going on here and also the code that's displaying the fiat rate is duplicated 😅 Can you refactor this to have clearly named boolean values to describe the different situations and extract the duplicate code to a shared var?

src/app/ui-components/selectFeeRate/feeSelectPopup.tsx
```diff
@@ -92,7 +91,11 @@ function FeeSelectPopup({
   baseToFiat,
   setFeeRate,
   getFeeForFeeRate,
+  absoluteBalance,
 }: Props) {
+  const stxInputExtractor = /[0-9]+[.]?[0-9]{0,6}/;
+  const inputValidator = absoluteBalance ? stxInputExtractor : /^[0-9]*$/;
```
victorkirov, line 97: sx specific code shouldn't live in this generic component. It also shouldn't rely on a non-stx related variable (absoluteBalance) to enable stx functionality. This should live in the outer, stx wrapper for this component.

The absoluteBalance arg also seems to be added in a non-generic way and the implementation seems to be specific to stx. I'm pretty sure this can also be removed in favor of the getFeeForFeeRate function returning undefined when there is not enough balance, as is done with the btc implementation.

src/app/ui-components/selectFeeRate/feeItem.tsx
```diff
@@ -171,17 +181,30 @@ function FeeItem({
...
+          {!feeRateUnits && fiat && (
+            <StyledP typography="body_medium_s" color={secondaryColor}>
+              <NumericFormat
+                value={fiat}
+                displayType="text"
+                prefix={`~${currencySymbolMap[fiatUnit]}`}
+                thousandSeparator
+                renderText={(value: string) => `${value} ${fiatUnit}`}
+              />
+            </StyledP>
```
victorkirov, line 199: This code is duplicated lower down with an inverse display logic. Please refactor to not have duplicate code.

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
```diff
@@ -0,0 +1,146 @@
...
+const RecipientInput = styled(Input)`
+  min-height: 125px;
+`;
+
+interface InputFeedback {
+  variant: FeedbackVariant;
+  message: string;
+}
+
+function Step1SelectRecipient({
```
victorkirov, line 43: Do we need to have StepX as a prefix? 😅 If we reorder the steps or add a step in between, this naming will be a chore to redo. Maybe better to remove the prefixes?

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
```diff
@@ -0,0 +1,146 @@
...
+  recipientAddress,
+  setRecipientAddress,
+  // recipientDomain,
```
victorkirov, line 46: no comment

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
```diff
@@ -0,0 +1,146 @@
...
+import Input from '@ui-library/input';
+import { FeedbackVariant } from '@ui-library/inputFeedback';
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import styled from 'styled-components';
+
+type StxRecipientScreenProps = {
+  recipientAddress: string;
+  setRecipientAddress: (address: string) => void;
+  // recipientDomain: string;
```
victorkirov, line 18: no comment

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
```diff
@@ -0,0 +1,146 @@
...
+    setInputFeedback(undefined);
+
+    if (associatedAddress !== '') {
+      setRecipientAddress(associatedAddress);
+      setRecipientDomain(recipient);
+      setInputFeedback([
+        { variant: 'checkmark', message: t('ASSOCIATED_ADDRESS') },
+        { variant: 'plainIndented', message: associatedAddress },
+      ]);
+    }
```
victorkirov, line 74: nit: avoid calling set state twice

```suggestion
    if (associatedAddress !== '') {
      setRecipientAddress(associatedAddress);
      setRecipientDomain(recipient);
      setInputFeedback([
        { variant: 'checkmark', message: t('ASSOCIATED_ADDRESS') },
        { variant: 'plainIndented', message: associatedAddress },
      ]);
    } else {
      setRecipientAddress(recipientAddress);
      setInputFeedback(undefined);
    }
```

victorkirov: I also don't really think this is needed: 
`setRecipientAddress(recipientAddress);`

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
```diff
@@ -0,0 +1,146 @@
...
+    setInputFeedback(undefined);
+
+    if (associatedDomain !== '') {
+      setRecipientDomain(associatedDomain);
+      setRecipientAddress(recipient);
+      setInputFeedback([
+        { variant: 'checkmark', message: t('ASSOCIATED_DOMAIN') },
+        { variant: 'plainIndented', message: associatedDomain },
+      ]);
+    }
```
victorkirov, line 90: nit: avoid calling setState twice

```suggestion
    if (associatedDomain !== '') {
      setRecipientDomain(associatedDomain);
      setRecipientAddress(recipient);
      setInputFeedback([
        { variant: 'checkmark', message: t('ASSOCIATED_DOMAIN') },
        { variant: 'plainIndented', message: associatedDomain },
      ]);
    } else {
      setRecipientDomain('');
      setInputFeedback(undefined);
    }
```

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
victorkirov: This useEffect and the other one both call  setRecipientDomain, setRecipientAddress and setInputFeedback. Since they are dependent on different other hooks (useBnsResolver and useBnsName), isn't there a race condition since both those hooks make api calls?

Maybe the 2 useEffects need to be combined into 1 which depends on associatedAddress and associatedDomain?

src/app/screens/sendStx/steps/Step3Confirm.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
+    navigate('/confirm-stx-tx', {
+      state: {
+        unsignedTx: unsignedSendStxTx,
+        fee,
+      },
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [unsignedSendStxTx]);
```
victorkirov, line 20: You don't really need any deps on this 😋

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
```diff
@@ -0,0 +1,146 @@
+import InputScreen from '@components/inputScreen';
+import { useBnsName, useBnsResolver } from '@hooks/queries/useBnsName';
+import useDebounce from '@hooks/useDebounce';
+// import { useBnsResolver } from '@hooks/queries/useBnsName';
+// import useDebounce from '@hooks/useDebounce';
```
victorkirov, line 5: no comment

src/app/screens/sendStx/stxAmountSelector.tsx
```diff
@@ -0,0 +1,250 @@
+import useWalletSelector from '@hooks/useWalletSelector';
```
victorkirov, line 1: There's a lot of duplication in this component and the btc one. Maybe we could extract a shared shell component which takes in values and functions which power the view and make btc and stx specific implementations which wrap the shell. It will be useful for reuse on brc-20, sip-10 and rune screens.

src/app/screens/sendStx/index.tsx
```diff
@@ -1,162 +1,143 @@
...
+    amountToSend,
+    stxMemo,
+    fee: previousFee,
+  } = location.state || {};
 
-    if (associatedAddress === stxAddress) {
-      setAddressError(t('ERRORS.SEND_TO_SELF'));
-      return false;
-    }
+  const [currentStep, setCurrentStep] = useState<Step>(stateAddress && amountToSend ? 1 : 0);
```
victorkirov, line 48: Would be better to use the enum values rather than 0 and 1.

Also, shouldn't we go to the amount selection screen if the address is set, regardless of whether an amount is set? 🤔

victorkirov: Should that not maybe be if address and the memo have values instead of the address and amount?

src/app/screens/sendStx/index.tsx
```diff
@@ -1,162 +1,143 @@
...
+  switch (currentStep) {
+    case Step.SelectRecipient:
+      return (
+        <SendLayout selectedBottomTab="dashboard" onClickBack={handleBackButtonClick}>
+          <Container>
+            <Step1SelectRecipientAndMemo
+              header={header}
+              recipientAddress={recipientAddress}
+              setRecipientAddress={setRecipientAddress}
+              // recipientDomain={recipientDomain}
```
victorkirov, line 100: no comment

victorkirov (changes requested):
Left a bunch of comments for consideration 😋

src/app/screens/sendBtc/recipientSelector.tsx
jordankzf: Sure, sounds good. [It'll solve this comment as well.](https://github.com/secretkeylabs/xverse-web-extension/pull/767#discussion_r1480977803)

src/app/screens/sendStx/index.tsx
```diff
@@ -1,162 +1,143 @@
...
+          <Container>
+            <Step2SelectAmount
+              header={header}
+              amount={amount}
+              setAmount={setAmount}
+              feeRate={feeRate}
+              setFeeRate={setFeeRate}
+              sendMax={sendMax}
+              setSendMax={setSendMax}
+              fee={feeRate}
```
abdulhaseeb4239, line 122: why do we need the `fee` prop when we already have `feeRate`? and we are passing the same value 🤔

src/app/screens/sendStx/steps/Step1SelectRecipient.tsx
jordankzf: The StepX prefix helped me a lot when troubleshooting / debugging, it's easier to go to the right file. I think it's worth the future chore of renaming it if we reorder the steps / add a step in-between. In my opinion, renaming is very trivial.

src/app/ui-components/selectFeeRate/feeSelectPopup.tsx
jordankzf: I think I did it this way because the BTC implementation could calculate the fees outside of the component with a transaction summary, do we have a similar analog for STX?

teebszet:
image

wrong font on the amount input fiat rate here

teebszet:
header bottom padding looks too large 
image

src/app/components/inputScreen/index.tsx
teebszet: I kinda agree with @victorkirov here. I get the intent is to abstract away some of the layout logic to make it consistent, but IMO a better abstraction is to store well-defined styled components into ui-library/common.styled.ts.

because in this case, i'm not sure if this abstracted layout will work for stx with the memo input.

src/app/ui-library/inputFeedback.tsx
```diff
@@ -1,23 +1,29 @@
...
 import Theme from 'theme';
 import { StyledP } from './common.styled';
 
-export type FeedbackVariant = 'info' | 'danger';
+export type FeedbackVariant = 'info' | 'danger' | 'plain' | 'plainIndented' | 'checkmark';
 
 const colors: Record<FeedbackVariant, string> = {
   info: Theme.colors.white_200,
   danger: Theme.colors.danger_light,
+  plain: Theme.colors.white_200,
```
teebszet, line 11: this should be 'explicative' to match zeroheight naming, and white_400
image

figma: 
image

src/app/ui-library/inputFeedback.tsx
```diff
@@ -1,23 +1,29 @@
...
+  plainIndented: Theme.colors.white_200,
```
teebszet, line 12: maybe this should be a `indentedNoIcon?: boolean` prop? 
so that any of the variants can replace the icon with an indent

teebszet: or just use 

noIcon?: boolean

and restyle this use where you're using it for bns names, because we don't want to bloat the ui-library with random exceptions that may not be used much in other places

src/app/ui-library/button.tsx
```diff
@@ -6,10 +6,13 @@ type ButtonVariant = 'primary' | 'secondary' | 'tertiary' | 'danger';
 
 const StyledButton = styled.button`
   width: 100%;
+  min-height: 42.5px;
```
teebszet, line 9: question: is this something that is needed for all buttons and their variations? in zeroheight it's height: 44px.

if it's only for an edge case usage, consider applying this min-height where you need it rather than enforcing it here

src/app/screens/sendStx/steps/Step2SelectAmount.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const { btcFiatRate, stxBtcRate, fiatCurrency, stxBalance, stxPublicKey, feeMultipliers } =
+    useWalletSelector();
+
+  const stxToFiat = (stx: string) =>
+    getStxFiatEquivalent(
+      stxToMicrostacks(new BigNumber(stx)),
+      new BigNumber(stxBtcRate),
+      new BigNumber(btcFiatRate),
+    )
+      .toNumber()
```
teebszet, line 92: question: is toNumber() necessary here?

src/app/screens/sendStx/steps/Step2SelectAmount.tsx
```diff
@@ -0,0 +1,233 @@
...
+            <SelectFeeRate
+              fee={fee}
+              feeUnits="STX"
+              feeRate={feeRate}
+              setFeeRate={setFeeRate}
+              baseToFiat={stxToFiat}
+              fiatUnit={fiatCurrency}
+              getFeeForFeeRate={getFeeForFeeRate}
+              feeRates={fees}
+              feeRateLimits={{ min: 0.000001, max: feeMultipliers?.thresholdHighStacksFee }}
```
teebszet, line 199: is this min/max logic stated somewhere in requirements?

src/app/screens/sendStx/steps/Step2SelectAmount.tsx
```diff
@@ -0,0 +1,233 @@
...
+        undefined,
+        selectedNetwork,
+      );
+      setFees({
+        low: Number(microstacksToStx(new BigNumber(low.fee)).toFixed(2)),
+        medium: Number(microstacksToStx(new BigNumber(medium.fee)).toFixed(2)),
+        high: Number(microstacksToStx(new BigNumber(high.fee)).toFixed(2)),
+      });
+      if (!fee)
+        setFeeRate(Number(microstacksToStx(new BigNumber(medium.fee)).toFixed(2)).toString());
```
teebszet, line 170: need to clarify with the requirements on what this logic should be. it looks to me that the notion requirements point to this section: https://www.notion.so/xverseapp/Speed-up-Stacks-transaction-Improve-fee-estimation-4455d27f8bdc4ebfbd94a16d7f18fe12?pvs=4#9509f0afd87d40eba297b357aa28763c

but there's no capping logic being applied here

teebszet:
I think the main points here are:

1. need to clarify and fix what the stx fee estimates should be (low, med, high)
2. need to fix the minor UI issues

3. need to think more carefully about how you're extending/duplicating the btc send max components, especially around the ui-components/selectFeeRate, as @victorkirov commented on. currently, the code to extend selectFeeRate for send stx feels a bit clunky

I think in order to keep this feature moving, let's focus on 1. and 2. to be ready for testing, and 3. can be a follow up task to include refactoring how best to share code between the various send flows, (including btc, stx, and other fungible token protocols)

src/app/screens/sendStx/steps/Step2SelectAmount.tsx
jordankzf: No, it's not documented. It's the same logic used for the current send stx implementation.

PR #774, opened 2024-01-25
Option to select xverse as default browser wallet in settings

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
When a user connects to an extension wallet, if there are multiple wallets implementing the sats connect interface, then the last wallet loaded would get all event messages. A user may want to force Xverse as the priority wallet to intercept these messages if the app using sats connect is using the default providers.

# 🔄 Changes
This adds a new setting toggle to set Xverse as the priority sats-connect wallet.

Note that the setting is stored in chrome storage so that it's available from the content-script directly without having to send messages around.

# 🖼 Screenshot / 📹 Video
image

# ✅ Review checklist


src/app/hooks/useChromeLocalStorage.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react';
+
+export const useChromeLocalStorage = <T extends unknown>(key: string, defaultValue?: T) => {
```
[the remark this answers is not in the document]
victorkirov: Ah, that's awesome. Will use that instead.

src/app/utils/chromeStorage.ts
```diff
@@ -19,35 +19,35 @@ class ChromeStorage {
...
           return reject(this.getError());
         }
-        return resolve(response[key]);
+        return resolve(response[key] ?? defaultValue);
       });
     });
   }
 
-  removeItem(key: string): Promise<any> {
+  removeItem(key: string): Promise<void> {
```
victorkirov, line 44: I tried changing this and the setItem one to boolean, but the seedVault stuff started to complain. Checking the rest of the code, nothing was actually using the return value, so I removed it.

src/app/utils/chromeStorage.ts
```diff
@@ -19,35 +19,35 @@ class ChromeStorage {
...
           return reject(this.getError());
         }
-        return resolve(true);
+        return resolve();
       });
     });
   }
 
-  getItem<T = any>(key: string): Promise<T> {
+  getItem<T = any, D = undefined>(key: string, defaultValue?: D): Promise<T | D> {
```
victorkirov, line 33: By default, the storage interface returns an empty object if the key is not present, resulting in a response of undefined. The Seed Vault expects null though, so I had to add this default value.

m-aboelenein (approved):

m-aboelenein:
> @m-aboelenein merged before it was tested!

I tested this one, and the change is pretty minimal we can retest with the RC

PR #779, opened 2024-01-26
Ensure correct ordinal is shown in ordinal detail view

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
This is a quick fix and first step to removing the selected NFT reducer. It ensures that the ordinal detail screen only shows the ordinal that the user selected.

# 🔄 Changes
Adds a check to ensure that the prepopulated ordinal data matches the selected ordinal.

To test:
1. Open the gallery page
2. Click on a collection
3. Click on an ordinal within the collection
4. Go back to the home screen from the bottom menu 
5. Go to the gallery view
6. Click on a different ordinal that is not part of a collection

The current prod version would show the first ordinal selected instead of the one you want to view. The new version will show the correct one.

# ✅ Review checklist

### DIFF
--- a/src/app/screens/ordinalDetail/useOrdinalDetail.ts
+++ b/src/app/screens/ordinalDetail/useOrdinalDetail.ts
@@ -2,7 +2,6 @@ import { useGetUtxoOrdinalBundle } from '@hooks/queries/ordinals/useAddressRareS
 import useInscriptionCollectionMarketData from '@hooks/queries/ordinals/useCollectionMarketData';
 import useAddressInscription from '@hooks/queries/ordinals/useInscription';
 import usePendingOrdinalTxs from '@hooks/queries/usePendingOrdinalTx';
-import useNftDataSelector from '@hooks/stores/useNftDataSelector';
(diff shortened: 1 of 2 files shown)


src/app/screens/ordinalDetail/useOrdinalDetail.ts
```diff
@@ -23,8 +22,7 @@ export default function useOrdinalDetail() {
   const { ordinalsAddress, network, selectedAccount, hasActivatedRareSatsKey } =
     useWalletSelector();
   const { id } = useParams();
-  const { selectedOrdinal } = useNftDataSelector();
-  const { data: ordinalData, isLoading } = useAddressInscription(id!, selectedOrdinal);
+  const { data: ordinalData, isLoading } = useAddressInscription(id!);
```
victorkirov, line 25: The ID is always populated as it's part of the route. In the final proper fix, we can also remove the second argument for useAddressInscription as it would just get the data from the fetch cache.

teebszet (approved):

PR #791, opened 2024-02-05
[ENG-3193] refactor: Refactor away from selectedOrdinal

terencehh:
# 🔘 PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Enhancement
- [x] Code style update (formatting, local variables)
- [X] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

# 📜 Background
There is a state bug where `selectedOrdinal` has improper state when on multiple tabs (it overwriting the current `selectedOrdinal` on the page). This PR introduces a fix by removing state management for `selectedOrdinal`, and instead always opting to look up the ordinal using the `ordinalId` in URL params, and finding the selected ordinal via react-query hooks (with caching)

Issue Link: [ENG-3193](https://linear.app/xverseapp/issue/ENG-3193/refactor-away-from-selectedordinal-and-selectedraresat)
Context Link (if applicable):

# 🔄 Changes
- removed `selectedOrdinal` state actions, reducers, and setters
- updated all `selectedOrdinal` usage to call react-query hook which retrieves ordinal details via the `ordinalId` supplied in URL params

Impact:
- fixes state management bugs when on multiple tabs in send ordinals flow (sendOrdinal & confirmOrdinalTransaction screens)
- indirectly improves SEO since URL contains data now

# 🖼 Screenshot / 📹 Video

Send/Confirm Ordinals Screen Bug video

https://github.com/secretkeylabs/xverse-web-extension/assets/30040178/4699cbe7-4718-4a61-84b0-cdaa313a1412

Send/Confirm Ordinals Screen Fix video

https://github.com/secretkeylabs/xverse-web-extension/assets/30040178/ca6f21dc-aecf-4a55-943a-679429a84938

Recover Funds/Ordinals Screen Bug video

https://github.com/secretkeylabs/xverse-web-extension/assets/30040178/d010e5d2-b53c-43b3-a35a-4594a583e6c9

Recover Funds/Ordinals Screen Fix video

https://github.com/secretkeylabs/xverse-web-extension/assets/30040178/96f1a5ec-4db0-48b2-9b91-77d452f00f4e

# ✅ Review checklist
Please ensure the following are true before merging:

- [x] Code Style is consistent with the project guidelines.
- [x] Code is readable and well-commented.
- [x] No unnecessary or debugging code has been added.
- [x] Security considerations have been taken into account.
[...]

PR #793, opened 2024-02-07
[ENG-3193] refactor: Refactor away from selectedSatBundle

terencehh:
# 🔘 PR Type

To be merged after #791.

- [x] Bugfix
- [ ] Enhancement
- [x] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background

This PR directly extends off the [previous PR](https://github.com/secretkeylabs/xverse-web-extension/pull/791), implementing additional state management fixes to `selectedSatBundle`.

context: when on multiple tabs (bundle data will overwrite the previous data on the old tab). This PR introduces a fix by removing state management for selectedSatBundle by supplying a `output (txid+vout)` to the `rare-sats-bundle` route, and calling react-query-hooks (with caching) to retrieve bundle data instead of relying on state management.

Issue Link: [ENG-3193](https://linear.app/xverseapp/issue/ENG-3193/refactor-away-from-selectedordinal-and-selectedraresat)
Context Link (if applicable):

# 🔄 Changes
- remove `selectedSatBundle` state actions, reducers, and setters
- updated all `selectedSatBundle` usage to call react-query hook which retrieves bundle details via the output (txid + vout) supplied in URL params

Impact:
- fixes state management bugs when on multiple tabs when navigating to Rare Sats Bundle page from Collectibles screen, or when navigating to Rare Sats Bundle page from Confirm Ordinal Screen / Ordinal Detail Screen
- indirectly improves SEO since URL contains data now

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


src/app/hooks/queries/ordinals/useAddressRareSats.ts
```diff
@@ -67,15 +67,31 @@ export const useGetUtxoOrdinalBundle = (
   });
 
   const bundle = data?.txid ? mapRareSatsAPIResponseToBundle(data) : undefined;
-  const inscriptionRange = bundle?.satRanges.find((range) =>
-    range.inscriptions.some((inscription) => inscription.inscription_number === ordinalNumber),
-  );
+
+  if (ordinalNumber) {
```
terencehh, line 71: not sure about this change, but ive noticed that we need this because when on `confirmOrdinalTransaction` screen, need to return data correctly when detecting to send an ordinal, or a satBundle

src/app/screens/confirmOrdinalTransaction/index.tsx
```diff
@@ -53,12 +52,24 @@ function ConfirmOrdinalTransaction() {
     fee = BigNumber(fee);
   }
   const { id } = useParams();
-  const { data: selectedOrdinal } = useAddressInscription(id!);
-  const { setSelectedSatBundleDetails } = useSatBundleDataReducer();
+  const [txid, vout] = id!.split(':');
```
teebszet, line 55: question: does this break the send ordinal flow?

src/app/screens/confirmOrdinalTransaction/index.tsx
```diff
@@ -53,12 +52,24 @@ function ConfirmOrdinalTransaction() {
...
+  const { data: selectedOrdinal } = useAddressInscription(txid!);
```
teebszet, line 56: i think this breaks the existing identifier protocol for ordinals inscriptions.

we still want the identifier of an ordinal transaction to be the inscriptionId like 
`const inscriptionId = "c71b7653ed0be8503851f028759921223f2c2d6f52cd39aebedcd0ddbf9b8962i0";`

teebszet: https://docs.ordinals.com/inscriptions.html#inscription-ids

src/app/screens/sendRareSat/index.tsx
```diff
@@ -135,7 +134,7 @@ function SendOrdinal() {
 
   useEffect(() => {
     if (data) {
-      navigate(`/nft-dashboard/confirm-ordinal-tx/${selectedSatBundle?.txid}`, {
+      navigate(`/nft-dashboard/confirm-ordinal-tx/${output}`, {
```
teebszet, line 137: I think this needs more thought..

src/app/screens/confirmOrdinalTransaction/index.tsx
```diff
@@ -53,12 +52,24 @@ function ConfirmOrdinalTransaction() {
...
+    isPartOfABundle,
+    ordinalSatributes,
+  } = useGetUtxoOrdinalBundle(
+    selectedOrdinal?.output || id,
+    hasActivatedRareSatsKey,
+    selectedOrdinal?.number,
+  );
+
+  const holdsRareSats = ordinalSatributes?.length > 0;
+
```
teebszet, line 72: I wonder if there's a cleaner way for this route to be able to handle both the send ordinal flow and send rare sat flow

teebszet (changes requested):
needs a cleaner solution to support the two use cases for confirmOrdinalTransaction

